PR #2, opened 2016-01-05
Multiples changes

fsaintjacques:
* code reformat
* usage of bool type
* usage of const keyword
* usage of __attribute(align) and fixed array for bitset_t


lemire:
PRs are fine but don't forget that you have write access to the repo.

PR #3, opened 2016-01-12
Cleanup code by uniformizing names

fsaintjacques:
I'm not directly pushing this one in master due to the merge-mess it might create with pending commits on local repos, @lemire I'll wait for your ok before merging this.

It compiles and passes tests for both AVX and NONAVX versions.


lemire:
Ok. It won't mess with my code but could clash with @nkurz let use see.

nkurz:
Go ahead and push what you'd like.  My changes are all messy test code, and
will require significant editing anyway.

PR #6, opened 2016-02-09
Cleanup array

fsaintjacques:
I'd like one or more pair of eyes to verify this set of changes.


lemire:
Will review. Soon. Not now.

include/util.h
```diff
@@ -1,58 +1,66 @@
-
 #ifndef UTIL_H
- 
 #define UTIL_H
 
+#define clamp(value, min, max) \
```
fsaintjacques: These 3 macros are not side-effect free, I'll need to transform them in static inline functions.

lemire:
Sorry about the delay. I have been dealing with other non-software issues. ;-)

include/util.h
```diff
@@ -1,58 +1,66 @@
...
+    ((value < min) ? min : (value > max) ? max : value)
 
-int32_t binarySearch(uint16_t* source, int32_t n, uint16_t target);
-
+#define roar_min(a, b) (a < b) ? a : b;
+#define roar_max(a, b) (a < b) ? b : a;
 
-int32_t advanceUntil(uint16_t * array, int32_t pos, int32_t length, uint16_t min);
```
lemire: Might be useful to document that ```advanceUntil``` is possibly implemented as an exponential search.

src/containers/array.c
```diff
@@ -12,799 +12,357 @@
...
+    } else {
+        free(array);
+        container->array = malloc(new_capacity * sizeof(uint16_t));
     }
+
     // TODO: handle the case where realloc fails
-    if(arr->array == NULL) {
-    	printf("Well, that's unfortunate. Did I say you could use this code in production?\n");
-    }
+    assert(container->array != NULL);
```
lemire: The assert is fine here, but, to be fair, it is not what one would want in production (not that we should worry about that here).

lemire:
This all looks fine to me. I recommend merging.

include/util.h
fsaintjacques: I didn't touch advanceUntil, this is an artefact from auto-formating.

lemire: I was aware of that, I was just making a general comment as much for myself as for others.

PR #19, opened 2016-03-17
Fixed issue with checking length of incorrect bitmap

deepakm:
A minor fix for in_place or, length2 was getting compared twice.


lemire:
Looks good.

Issue #29, opened 2016-05-06
Implement and test copy-on-write


lemire:
In progress at https://github.com/RoaringBitmap/CRoaring/tree/performance/cow

lemire:
c.c. @owenkaser

lemire:
Implemented in this PR: https://github.com/RoaringBitmap/CRoaring/pull/30

lemire:
Done and tested.

PR #30, opened 2016-05-08
Implemented copy-on-write (CoW) at the container level. 

lemire:
This seems to pass our unit tests and it seems to generally help our performance without doing any harm.

Basically, it helps the performance when you would otherwise have to clone containers needlessly. It is not just the copying, but also the memory usage, the memory allocation and so forth. How often this happens depends very much on the data. However, avoiding even just one container copy can maybe be worth it.

To see how much it can help, here is the number of cycles per input value for successive unions in the census1881 data set. 

> Before (no CoW): 1.24 
> After (with CoW):  0.49 

To my surprise, there is at least one instance where we can double the performance with CoW.

It does, however, slightly increase the code complexity. Maybe surprisingly, however, it is not so difficult to implement. It is possible however that I introduced memory leaks and so forth. Thread safety might be diminished as well, maybe. I don't know.

Before merging this with ``master``, I'd like for at least on fresh pair of eyes to review it. Indeed, if we adopt this, then it will have long-term consequences regarding maintenance and future work.

It is likely, also, that I have let bugs slip.

c.c. @owenkaser  @fsaintjacques


Ezibenroc:
Hi,

I coded a small Python wrapper over your library (see [here](https://github.com/Ezibenroc/PyRoaringBitMap/tree/master)).
I noticed that this new feature break my unit tests, so I wanted to tell you.

Here is a minimal piece of code that shows the bug:
```c
void mytest() {
    roaring_bitmap_t *bm1 = roaring_bitmap_create();
    roaring_bitmap_add(bm1, 3);
    roaring_bitmap_t *bm2 = roaring_bitmap_copy(bm1);
    roaring_bitmap_add(bm2, 4);
    roaring_bitmap_add(bm1, 5);
    printf("card(bm1) = %u\n", (unsigned)roaring_bitmap_get_cardinality(bm1));
    printf("card(bm2) = %u\n", (unsigned)roaring_bitmap_get_cardinality(bm2));
    printf("bm1 = ");
    roaring_bitmap_printf(bm1);
    printf("\n");
    printf("bm2 = ");
    roaring_bitmap_printf(bm2);
    printf("\n");
    roaring_bitmap_free(bm1);
    roaring_bitmap_free(bm2);
}
```

It prints the following:
```
card(bm1) = 1
card(bm2) = 1
bm1 = {3}
bm2 = {3}
```
It seems that after the copy, we cannot add any new element in the two bitmaps.
Note that if we copy an empty bitmap (i.e. remove the line `roaring_bitmap_add(bm1, 3);`), everything seems to work fine.

lemire:
@Ezibenroc 

Hopefully fixed with this commit : 

https://github.com/RoaringBitmap/CRoaring/pull/30#issuecomment-217704058

lemire:
Please keep hunting for bugs. I am sure that are more... ;-)

lemire:
This is meant to satisfy issue 29 https://github.com/RoaringBitmap/CRoaring/issues/29

lemire:
Note: I am not working on this code. If someone wants to hack the code, please do.

c.c. @owenkaser

lemire:
Ah. This code had leaks. With the leaks fixed, we get out-of-bound accesses which suggests that the code has problems.

c.c. @owenkaser

lemire:
With commit https://github.com/RoaringBitmap/CRoaring/pull/30/commits/a60350d03fd817cb586e6dcac6a921c92d6f19ce  there is no obvious data corruption.

lemire:
Propose change to the API so that copy-on-write is optional...

https://github.com/RoaringBitmap/CRoaring/pull/30/commits/4e5b06a07a42e3845e9170538f818dcc7986d75a

This should make concurrency easier.

lemire:
@fsaintjacques 

Can you have a quick look, at least at ``roaring.h``?

There is an API change and I'd like someone to review it before I merge. I think that the code is sound and well motivated... it is a nice feature to have ("almost free copies"), but I am concerned with usability.

include/containers/containers.h
```diff
@@ -34,18 +121,52 @@ static inline const char *get_container_name(uint8_t typecode) {
...
+                return "unknown";
+        	}
+        	break;
+        default:
+            assert(false);
+            __builtin_unreachable();
+            return "unknown";
+    }
+    __builtin_unreachable();
+}
```
brentp: if you start with 
```
if typecode == SHARED_CONTAINER_TYPECODE {
    typecode = (shared_container_t *)container)->typecode)
}
```
you can remove some redundancy.

brentp:
could you have it so that cow is configurable at the roaring_bitmap_t level? e.g. rbm->cow = 1.
then the function signatures could remain unchanged in a lot of cases.

lemire:
@brentp  You are right. It would be nicer.

include/containers/containers.h
lemire: Well, this would not allow to distinguish shared from unshared containers.

fsaintjacques:
Agreedo. @lemire I'll try to find time tonight to review this change.

lemire:
@fsaintjacques 

I'll try to implement @brentp's idea today as I think it is really much nicer than what I've got.

lemire:
@fsaintjacques  I can squash the commits if you want.

fsaintjacques:
No need to squash.

lemire:
@fsaintjacques 

I am going to propose that we merge this, whether or not you have time to review it in the near future, because we are quickly going to be chasing our tail, with new features making it into master without COW... 

Thoughts?

fsaintjacques:
Sorry for the lack of answers, I'm fairly busy.

lemire:
@fsaintjacques  That's ok.

PR #34, opened 2016-05-22
Bitmap constructor from a range

Ezibenroc:
This pull request implements the following function:

```c
roaring_bitmap_t *roaring_bitmap_from_range(uint32_t min, uint32_t max, uint32_t step);
```

It creates a bitmap with values in `[min, min+step, min+2*step, ..., min+k*step]` such that `k*step<max` and `(k+1)*step>=max` (similar to what does the Python `range`).

To do this efficiently, two tricks are used:
* When the container is an array, we first construct an empty array allocated with the required size, then we append all the values.
* When the container is a bitset and `step` is a divisor of 64, all the cells of the bitsets are the same, so we just compute it once.

For `min=0, max=1000000000, step=1`, we get a speedup of about 200 with this solution, in comparison with the naive idea to use `roaring_bitmap_add` for every value of the range.

There is certainly other optimizations to do.

### DIFF
--- a/include/containers/containers.h
+++ b/include/containers/containers.h
@@ -61,7 +61,7 @@ static inline int container_get_cardinality(const void *container,
 
 /*  Create a container with all the values between min and max (included) at a
     distance k*step from min. */
-static inline void *container_from_range(int *type, uint16_t min, uint16_t max,
+static inline void *container_from_range(uint8_t *type, uint16_t min, uint16_t max,
                                                   uint16_t step) {
     uint16_t size = (max-min)/step;
     if(size < 4096) { // array container
--- a/include/containers/containers.h
+++ b/include/containers/containers.h
@@ -59,6 +59,25 @@ static inline int container_get_cardinality(const void *container,
     return 0;  // unreached
 }
 
+/*  Create a container with all the values between min and max (included) at a
+    distance k*step from min. */
+static inline void *container_from_range(int *type, uint16_t min, uint16_t max,
+                                                  uint16_t step) {
+    uint16_t size = (max-min)/step;
+    if(size < 4096) { // array container
+        *type = ARRAY_CONTAINER_TYPE_CODE;
+        array_container_t * array = array_container_create_given_capacity(size);
+        array_container_add_from_range(array, min, max, step);
+        return array;
+    }
+    else { // bitset container
+        *type = BITSET_CONTAINER_TYPE_CODE;
+        bitset_container_t *bitset = bitset_container_create();
+        bitset_container_add_from_range(bitset, min, max, step);
+        return bitset;
+    }
+}
+
 /**
  * "repair" the container after lazy operations.
  */
--- a/src/roaring.c
+++ b/src/roaring.c
@@ -55,6 +55,16 @@ roaring_bitmap_t *roaring_bitmap_of(size_t n_args, ...) {
     return answer;
 }
 
+roaring_bitmap_t *roaring_bitmap_from_range(uint32_t min, uint32_t max, uint32_t step) {
+    if(step == 0)
+        return NULL;
+    roaring_bitmap_t *answer = roaring_bitmap_create();
+    for(uint32_t value = min ; value < max ; value += step) {
+            roaring_bitmap_add(answer, value);
+    }
+    return answer;
+}
+
 void roaring_bitmap_printf(const roaring_bitmap_t *ra) {
     printf("{");
     for (int i = 0; i < ra->high_low_container->size; ++i) {
--- a/src/roaring.c
+++ b/src/roaring.c
@@ -55,13 +55,41 @@ roaring_bitmap_t *roaring_bitmap_of(size_t n_args, ...) {
     return answer;
 }
 
+static inline int32_t minimum(uint32_t a, uint32_t b) { return (a < b) ? a : b; }
+
+static inline int32_t next_container_bound(uint32_t value) {
+    return (1<<16)*(value/(1<<16)) + (1<<16);
+}
+
 roaring_bitmap_t *roaring_bitmap_from_range(uint32_t min, uint32_t max, uint32_t step) {
     if(step == 0)
         return NULL;
     roaring_bitmap_t *answer = roaring_bitmap_create();
-    for(uint32_t value = min ; value < max ; value += step) {
-            roaring_bitmap_add(answer, value);
+    if(max <= min)
+        return answer;
+    if(step >= (1<<16)) {
+        for(uint32_t value = min ; value < max ; value += step) {
+                roaring_bitmap_add(answer, value);
+        }
+        return answer;
     }
+    uint32_t min_tmp = min, max_tmp = next_container_bound(min);
+    do {
+        uint16_t key = min_tmp/(1<<16);
+        uint32_t real_max = minimum(max, max_tmp)-1;
+        uint32_t size = (real_max-min_tmp)/step;
+        if(size < 4096) { // array container
+            array_container_t * array = array_container_create_given_capacity(size);
+            array_container_add_from_range(array, min_tmp-key*(1<<16), real_max-key*(1<<16), (uint16_t)step);
+            ra_append(answer->high_low_container, key, array, ARRAY_CONTAINER_TYPE_CODE);
+        }
+        else { // bitset container
+            // // TODO
+            return NULL;
+        }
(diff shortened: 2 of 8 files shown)


tests/toplevel_unit.c
```diff
@@ -120,6 +120,41 @@ void test_example() {
...
 
+bool check_bitmap_from_range(uint32_t min, uint32_t max, uint32_t step) {
+    roaring_bitmap_t *result = roaring_bitmap_from_range(min, max, step);
+    assert_non_null(result);
+    roaring_bitmap_t *expected = roaring_bitmap_create();
+    assert_non_null(expected);
+    for(uint32_t value = min ; value < max ; value += step) {
+        roaring_bitmap_add(expected, value);
+    }
+    bool is_equal = roaring_bitmap_equals(expected, result);
```
fsaintjacques: You can replace the following 5 lines with a simple

`assert_true(expr);`

tests/toplevel_unit.c
```diff
@@ -120,6 +120,41 @@ void test_example() {
...
+        fprintf(stderr, "[ERROR] check_bitmap_from_range(%u, %u, %u)\n",
+            (unsigned)min, (unsigned)max, (unsigned)step);
+    }
+    return is_equal;
+}
+
+void test_bitmap_from_range() {
+    assert_true(roaring_bitmap_from_range(1, 10, 0) == NULL); // undefined range
+    bool no_error = true;
+    no_error = check_bitmap_from_range(5, 1, 3) && no_error; // empty range
```
fsaintjacques: Same here, a simple `assert_true` should suffice.

src/roaring.c
```diff
@@ -55,6 +55,39 @@ roaring_bitmap_t *roaring_bitmap_of(size_t n_args, ...) {
...
+
+static inline int32_t next_container_bound(uint32_t value) {
+    return (1<<16)*(value/(1<<16)) + (1<<16);
+}
+
+roaring_bitmap_t *roaring_bitmap_from_range(uint32_t min, uint32_t max, uint32_t step) {
+    if(step == 0)
+        return NULL;
+    roaring_bitmap_t *answer = roaring_bitmap_create();
+    if(max <= min)
```
fsaintjacques: Does this follow the behaviour in python? From my perspective, I'd prefer a `NULL`.

include/containers/containers.h
```diff
@@ -59,6 +59,25 @@ static inline int container_get_cardinality(const void *container,
     return 0;  // unreached
 }
 
+/*  Create a container with all the values between min and max (included) at a
+    distance k*step from min. */
+static inline void *container_from_range(int *type, uint16_t min, uint16_t max,
```
fsaintjacques: If I recall, `type` should be a `uint8_t`?

lemire:
Seems like a great idea.

fsaintjacques:
I'd suggest to have a `roaring_bitmap_add_range ` and then a `roaring_bitmap_from_range` that calls the first method.

tests/toplevel_unit.c
Ezibenroc: The idea of this code was to run all the tests related to `roaring_bitmap_from_range` and output a message for every eventual error. The goal is to ease the debugging.

If I modify the code like you say, then : 
* The test stops after one error is detected.
* We don't know which sub-test failed (i.e. the values of `min`, `max` and `step`).

Do you have any idea on how to avoid these issues? I think the second is the most important.

src/roaring.c
Ezibenroc: Yes, in Python `range(10, 3)` is valid code, which is simply an empty range.

I find this empty range nice, but it may hide integer overflows and thus produce some nasty bugs. So yes, returning `NULL` may be better. Do you have other concerns about this?

Ezibenroc:
> I'd suggest to have a roaring_bitmap_add_range and then a roaring_bitmap_from_range that calls the first method.

I am not sure it is worth it since it would complicate the code and make it less efficient.
For instance, when the container is an array, we cannot suppose anymore that it is empty, so we cannot simply append the values to its end.
We also cannot guess in advance the sizes of the resulting containers. Therefore, we need to compute them after (which is an important overhead for bitsets) and container conversions are needed in some cases.

lemire:
@Ezibenroc 

This is cool stuff. 

1. I think @fsaintjacques  has a point regarding the container type, you use an ``int`` type, in a way that appears safe, but the convention in the code is that we use an ``uint8_t`` type. Would you consider doing this minor fix?

2. I think that the arguments regarding how to set up the unit tests are minor. No?

3. I think your code is slightly inconsistent regarding whether it returns ``NULL`` (when ``step = 0``) or an empty bitset. I think ``NULL`` might be nicer since, some of the time you will be returning ``NULL``, might as well return ``NULL`` whenever something is off. Would you considering making this change?

tests/toplevel_unit.c
fsaintjacques: While I agree that in production code, having the debug info is relevant, when in test suites, reproducing with GDB is trivially easy and thus doesn't need this type of details.

fsaintjacques:
The arguments with the tests are more of an aesthetic preferences more than an objection.

tests/toplevel_unit.c
lemire: @fsaintjacques 

Here I would think that printing out what is failing exactly can speed up debugging considerably. 

And where is the harm in @Ezibenroc's code?

tests/toplevel_unit.c
lemire: @Ezibenroc 

>  We don't know which sub-test failed (i.e. the values of min, max and step).

That part, I agree with... I think it is useful for the test to fail in a verbose way so that we know what failed right away.


> The test stops after one error is detected.

This is not necessarily a negative. 

@fsaintjacques 

> reproducing with GDB is trivially easy and thus doesn't need this type of details.

Running things in a debugger is slow and time consuming. If the test tells you exactly what is wrong, you can often immediately act and save yourself 15 minutes of fiddling with gdb.

Ezibenroc:
Alright, I have fixed the type and the returned value.

lemire:
@fsaintjacques  Are we good to merge? (with the convention that aesthetic preferences should probably not stop us)

lemire:
@Ezibenroc 

I did some cleaning and minor optimizations

https://github.com/RoaringBitmap/CRoaring/commit/cbf047ddbb6cc36347b6dea227b1c931e0db7ee2

One optimization (the use of run containers) can be significant in some instances.

Ezibenroc:
Great! It has even corrected a bug that I just found.

However, there is now a new bug. The following test fails:
```c
    roaring_bitmap_t *bm1 = roaring_bitmap_from_range(0, 1, 1);
    roaring_bitmap_t *bm2 = roaring_bitmap_from_range(0, 2, 1);
    assert_false(roaring_bitmap_equals(bm1, bm2));
```

lemire:
@Ezibenroc 

Let me have a look.

lemire:
@Ezibenroc  Fixed. https://github.com/RoaringBitmap/CRoaring/commit/daa1b4f67053162119f8af4cfbb94841d4f3fa53

lemire:
Goes without saying that nobody should be using this code in production (yet).

Ezibenroc:
There is also an issue when mixing range initialization and serialization. The following test fails:

```c
    roaring_bitmap_t *old_bm = roaring_bitmap_from_range(65520, 131057, 16);
    size_t size = roaring_bitmap_portable_size_in_bytes(old_bm);
    char *buff = malloc(size);
    roaring_bitmap_portable_serialize(old_bm, buff);
    roaring_bitmap_t *new_bm = roaring_bitmap_portable_deserialize(buff);
    assert_true(roaring_bitmap_equals(old_bm, new_bm));
    roaring_bitmap_free(old_bm);
    roaring_bitmap_free(new_bm);
    free(buff);
```

It works fine if you replace the range constructor by this:
```c
    roaring_bitmap_t *old_bm = roaring_bitmap_create();
    for(uint32_t value = 65520 ; value < 131057 ; value += 16)
        roaring_bitmap_add(old_bm, value);
```

lemire:
@Ezibenroc  Ok. I'll look at this right away.

lemire:
@Ezibenroc  Fixed https://github.com/RoaringBitmap/CRoaring/commit/9fed6877e1f595e7320d7b4c0347bb1a822c8833

PR #35, opened 2016-05-31
Fix roaring_bitmap_remove by avoiding a double free.

Ezibenroc:
The container was freed twice: in `roaring_bitmap_remove` and in `ra_remove_at_index`.
The double free was crashing the program with the simple test I added.

I also think there was a typo in `test_remove_from_copies_false`, I fixed it.

### DIFF
--- a/src/roaring.c
+++ b/src/roaring.c
@@ -190,7 +190,6 @@ void roaring_bitmap_remove(roaring_bitmap_t *r, uint32_t val) {
             ra_set_container_at_index(r->high_low_container, i, container2,
                                               newtypecode);
         } else {
-            container_free(container2, newtypecode);
             ra_remove_at_index(r->high_low_container, i);
         }
     }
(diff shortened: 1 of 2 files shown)


lemire:
I'll merge and check.

Ezibenroc:
The following also crashes:
```c
void test_strange_bug() {
    roaring_bitmap_t *bm = roaring_bitmap_create();
    for(uint32_t value = 33057 ; value < 147849 ; value += 8) {
        roaring_bitmap_add(bm, value);
    }
    for(uint32_t value = 33057 ; value < 147849 ; value += 8) {
        roaring_bitmap_remove(bm, value);
    }
}
```

(Valgrind says there is some invalid write in `bitset_extract_setbits_uint16`)

lemire:
Ok. I merged, but it is no longer passing unit tests, maybe because you fixed the unit test to actually test what we want to test. So there is still a problem.

lemire:
Turns out that I had left a memory leak in the unit test. Fixed. Let me have a look at your new test.

lemire:
Yes. There is a problem. You are right.

lemire:
Fixed

Issue #33, opened 2016-05-13
Implement rank queries


lemire:
Implement `roaring_bitmap_rank`: given a value x, find how many values in the bitmaps are in [0,x).

lemire:
See PR https://github.com/RoaringBitmap/CRoaring/pull/38 by @Ezibenroc

lemire:
Select part implemented by @Ezibenroc 

We have to do the rank part.

lemire:
Rank part done in v0.2.15

PR #38, opened 2016-06-06
Implement get_element_of_rank.

Ezibenroc:
The goal here is to add this function:

```c
bool roaring_bitmap_get_element_of_rank(roaring_bitmap_t *ra, uint32_t rank, uint32_t *element);
```

It retrieves an element of given rank in the roaring bitmap.
To do so, we simply iterate on all the containers until we find the desired rank (or we reach the end if the rank is greater than the bitmap's size).

Some factorization of the code could be made, particularly in the tests. For instance, we could have a function returning a randomly generated bitmap. We could also factorize the code for some container tests into a generic file (e.g. `container_unit.c`).
But I think this should be the object of an other pull request.

Future work regarding rank queries can be:
* Implement rank range queries (e.g. to get a new bitmap containing all the elements between the 10th element and the 27th).
* Implement rank queries in the other way (we have an element and we want to know its rank).

### DIFF
--- a/include/containers/array.h
+++ b/include/containers/array.h
@@ -213,7 +213,7 @@ bool array_container_equals(array_container_t *container1,
  * accordingly.
  * Otherwise, it returns false and update start_rank.
  */
-static inline bool array_container_select(array_container_t *container,
+static inline bool array_container_select(const array_container_t *container,
                                           uint32_t *start_rank, uint32_t rank, uint32_t *element) {
     int card = array_container_cardinality(container);
     if(*start_rank + card <= rank) {
--- a/include/containers/array.h
+++ b/include/containers/array.h
@@ -213,7 +213,7 @@ bool array_container_equals(array_container_t *container1,
  * accordingly.
  * Otherwise, it returns false and update start_rank.
  */
-static inline bool array_get_element_of_rank(array_container_t *container,
+static inline bool array_container_select(array_container_t *container,
                                           uint32_t *start_rank, uint32_t rank, uint32_t *element) {
     int card = array_container_cardinality(container);
     if(*start_rank + card <= rank) {
--- a/include/containers/array.h
+++ b/include/containers/array.h
@@ -207,4 +207,23 @@ static inline int32_t array_container_size_in_bytes(
 bool array_container_equals(array_container_t *container1,
                             array_container_t *container2);
 
+/**
+ * If the element of given rank is in this container, supposing that the first
+ * element has rank start_rank, then the function returns true and sets element
+ * accordingly.
+ * Otherwise, it returns false and update start_rank.
+ */
+static inline bool array_get_element_of_rank(array_container_t *container,
+                                          uint32_t *start_rank, uint32_t rank, uint32_t *element) {
+    int card = array_container_cardinality(container);
+    if(*start_rank + card <= rank) {
+        *start_rank += card;
+        return false;
+    }
+    else {
+        *element = container->array[rank-*start_rank];
+        return true;
+    }
+}
+
 #endif /* INCLUDE_CONTAINERS_ARRAY_H_ */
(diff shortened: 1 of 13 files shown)


lemire:
I think that this is part of https://github.com/RoaringBitmap/CRoaring/issues/33

To be reviewed.

lemire:
@Ezibenroc 

Here is what I think is the proper terminology...

1. A ``rank`` query is such that your return the number of elements that are smaller or equal to x (rank(infinity) would be the cardinality). 
2. A ``select`` query is such that your return the jth value stored in this bitmap. (I think you would say "the element having rank j".)

So I think that what you implemented is a ``select`` not, a ``rank``. So I think that a less confusing name for your function would be ``roaring_bitmap_select``...

See wikipedia....

> In computer science, a selection algorithm is an algorithm for finding the kth smallest number in a list or array (https://en.wikipedia.org/wiki/Selection_algorithm)

I think that's what you implemented, and it is typically called a ``selection``. Now, you might reasonably argue that your English is proper, you "return the element having a given rank". But then how are we going to call the function that does a ``rank``? ``roaring_bitmap_get_rank_of_element``. Won't that get confusing?

I think we should call the functions...

* ``roaring_bitmap_select``
* ``roaring_bitmap_rank``

Of course, this is semantics and we can and should debate these things until we are all happy.

Though your implementation looks good to my eyes, might I entice you to briefly review the Java version so that you can be sure your implementation is at least as nice and efficient? Here are the relevant parts (it is not long or complicated) : 

https://github.com/RoaringBitmap/RoaringBitmap/blob/master/src/main/java/org/roaringbitmap/RoaringBitmap.java#L1771-L1785

https://github.com/RoaringBitmap/RoaringBitmap/blob/master/src/main/java/org/roaringbitmap/ArrayContainer.java#L961-L964

https://github.com/RoaringBitmap/RoaringBitmap/blob/master/src/main/java/org/roaringbitmap/BitmapContainer.java#L1078-L1089

https://github.com/RoaringBitmap/RoaringBitmap/blob/master/src/main/java/org/roaringbitmap/RunContainer.java#L1950-L1961

Doing this has also the added benefit that the Java version has already a rank implementation, so we can "almost" get it for free.

fsaintjacques:
I agree with @lemire, this is proper nomenclature. See this amazing library:

https://github.com/simongog/sdsl-lite

Ezibenroc:
Sorry for this terminology misunderstanding, I will rename the functions to `roaring_bitmap_select`.

After a quick look to the Java code, I think I have done something similar but maybe more complicated.
I do not call `get_cardinality` on the containers from the `roaring_bitmap` function, since it implies some computations for the run container. If the element we are looking for is in the run container, we will iterate over it twice: the first time to compute its cardinality, the second time to get the element.
But maybe the overhead is negligible?

lemire:
@Ezibenroc 

Yes, I think your implementation is fine compared to the Java one, but it may yet be slightly buggy in the sense that you do not support lazy-on-write. Specifically, I think you are missing a line such as this one...
https://github.com/RoaringBitmap/CRoaring/blob/master/include/containers/containers.h#L170


More critically, your unit tests do not show a problem with copy-on-write whereas I think they show... I think you should add a test that starts like this...
```
    roaring_bitmap_t *bmtmp = roaring_bitmap_create();
    // populate bitmap here
    ...
    //
    // do a lazy copy...
    bm1->copy_on_write = true;
    roaring_bitmap_t *bm = roaring_bitmap_copy(bmtmp);
```

If I am right, you need to add a couple of lines here and there in your code.

Ezibenroc:
Ok, I forgot the copy-on-write functionality.

It should be fixed now.

lemire:
Looks good to me.

lemire:
Merging.

PR #40, opened 2016-06-20
Add UINT64_C and UINT32_C definitions to fix portability issues.

Ezibenroc:
This fixes the compilation on an old system.


lemire:
It seems it can do no harm but I am pretty sure that these macros are part of C99.

PR #41, opened 2016-06-24
Fix flip_inplace by removing a double free.

Ezibenroc:
The method `roaring_bitmap_flip_inplace` was causing a segmenntation fault when
a container of the bitmap had to be removed. It was due to a double free.

This PR fixes this by replacing two calls to `ra_remove_at_index_and_free` by calls
to `ra_remove_at_index`, in file `roaring.c`.
According to the tests, it does not introduce any memory leak, but we should check
carefully.


lemire:
@owenkaser  Given that this is your code... do you have time to review the PR?

owenkaser:
yes, I will review.

owenkaser:
The revised logic made sense and the bug escaped notice before only due to insufficient test cases.

PR #43, opened 2016-07-14
Fix include paths for library header files

thus:
I installed CRoaring the following way:
```
mkdir build && cd build
cmake ..
make
make install
```
But I keep running into the following error when including it using "roaring/roaring.h":
```
user@debian:~# gcc roaring-bitmap-test.c -lroaring
In file included from /usr/local/include/roaring/containers/containers.h:8:0,
                 from /usr/local/include/roaring/roaring_array.h:7,
                 from /usr/local/include/roaring/roaring.h:12,
                 from bitmap-test.c:1:
/usr/local/include/roaring/containers/array.h:13:25: fatal error: portability.h:
No such file or directory
 #include "portability.h"
                         ^
compilation terminated.
```
This error goes away if I add the include path for the CRoaring header files (gcc -I/usr/local/include/roaring [...]) when compiling. But, I think it would be better if it would just use the default OS include path instead of having to specify it ourself using compiler flags.

This pull request change the library header include paths. Your feedback is highly appreciated.

By the way, your project is awesome!


lemire:
@fsaintjacques  Want to have a look?

fsaintjacques:
This fix seems reasonable. As I mentioned previously, we should move all private headers into `src/` and divide `roaring.h` in 3 files: 

- roaring.h (main methods)
- roaring_io.h (io related methods)
- roaring_misc.h (various thing like stats, optimize rle, ...)

lemire:
@fsaintjacques @thus 

Ok. Let us merge.

thus:
Thanks :)

lemire:
@thus Thanks to you!

PR #61, opened 2016-09-17
Implement is_subset function, close #55

### LINKED ISSUE #55: Add a subset test
We can test if two bitmaps are equal, but we cannot test if one is included in the other.
This can be done by computing the intersection (or the union) of the two bitmaps, but I guess that this is not very efficient.

We should add two functions like this:

```c
// Return true if and only if ra1 ⊆ ra2
bool roaring_bitmap_included(roaring_bitmap_t *ra1, roaring_bitmap_t *ra2);
// Return true if and only if ra1 ⊊ ra2
bool roaring_bitmap_strictly_included(roaring_bitmap_t *ra1, roaring_bitmap_t *ra2);
```
We can easily compute one function with the other and `roaring_bitmap_equal`, but again this would not be very efficient.

Ezibenroc:
The main contribution of this pull request is the addition of the two functions: 

```c
/**
 * Return true if all the elements of ra1 are also in ra2.
 */
bool roaring_bitmap_is_subset(roaring_bitmap_t *ra1, roaring_bitmap_t *ra2);

/**
 * Return true if all the elements of ra1 are also in ra2 and ra2 is strictly greater
 * than ra1.
 */
bool roaring_bitmap_is_strict_subset(roaring_bitmap_t *ra1, roaring_bitmap_t *ra2);
```

They are implemented following the same spirit than `roaring_bitmap_equals`:

* The function `roaring_bitmap_is_subset` iterates on the containers of the two bitmaps. When two containers have a same key, they are compared.
* The comparison of these two containers is done in the generic function `container_is_subset`. It calls the right function according to the type of the two containers.
* Functions to compare containers are implemented as `bitset_container_is_subset`, `bitset_container_is_subset_run`, `run_container_is_subset_bitset`, `array_container_is_subset_bitset`, `array_container_is_subset_run`, `run_container_is_subset_array`, `array_container_is_subset` and `run_container_is_subset`.

Unit tests for these functions are in file `container_comparison_unit.c`. I also added the unit tests for `container_equals` since they were missing and they follow the same structure.

There is also a test in `toplevel_unit.c`.

----

**Future work**

We can still do optimization, for instance on functions `bitset_container_is_subset_run` and `run_container_is_subset_bitset`. The same kind of optimization can also be done on function `run_container_equals_bitset` which existed prior to this pull request.

### DIFF
--- a/amalgamation.sh
+++ b/amalgamation.sh
@@ -28,6 +28,7 @@ $SCRIPTPATH/include/roaring/containers/bitset.h
 $SCRIPTPATH/include/roaring/containers/run.h
 $SCRIPTPATH/include/roaring/containers/convert.h
 $SCRIPTPATH/include/roaring/containers/mixed_equal.h
+$SCRIPTPATH/include/roaring/containers/mixed_subset.h
 $SCRIPTPATH/include/roaring/containers/mixed_andnot.h
 $SCRIPTPATH/include/roaring/containers/mixed_intersection.h
 $SCRIPTPATH/include/roaring/containers/mixed_negation.h
--- a/include/roaring/containers/array.h
+++ b/include/roaring/containers/array.h
@@ -205,6 +205,12 @@ bool array_container_equals(array_container_t *container1,
                             array_container_t *container2);
 
 /**
+ * Return true if container1 is a subset of container2.
+ */
+bool array_container_is_subset(array_container_t *container1,
+                            array_container_t *container2);
+
+/**
  * If the element of given rank is in this container, supposing that the first
  * element has rank start_rank, then the function returns true and sets element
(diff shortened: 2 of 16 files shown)


lemire:
Thanks. Excellent.

I looked at the code and it looks excellent. 

@Ezibenroc  Do you recommend an immediate merge?

Ezibenroc:
@lemire I think that the remaining optimizations can be done later, we should not delay the merge because of this.

We could also write other tests. The current ones are very generic and we certainly do not cover all corner cases. But writing more extensive tests will be tedious.

So, unless you see any easy optimization or test, yes I think we should merge.

lemire:
Ok.

PR #70, opened 2016-12-26
Fix memory leak in roaring_bitmap_or and roaring_bitmap_xor.

Ezibenroc:
When one of the two bitmaps is empty, the functions return the other bitmap, so we should not allocate a new one.


lemire:
Wow. Our valgrind-based tests should have revealed this.

lemire:
LGTM but I'll be adding corresponding tests.

gssiyankai:
Have you considered using smart pointers ? This would have prevented memory leaks with a small overhead.

lemire:
@gssiyankai  This is a C library so I don't think smart pointers are an option. Not out of the box.

Of course, we could use automated memory management techniques, but these tend to impact how interoperable the code is...

PR #72, opened 2017-01-05
Made changes allowing headers to be included in multiple compilation units

ksajme:
Including the C++ amalgamation in multiple compilation units causes some linker errors due to a few missing "inline"s.

### DIFF
--- a/cpp/roaring.hh
+++ b/cpp/roaring.hh
@@ -506,11 +506,11 @@ public:
 };
 
 
-RoaringSetBitForwardIterator Roaring::begin() const {
+inline RoaringSetBitForwardIterator Roaring::begin() const {
       return RoaringSetBitForwardIterator(*this);
 }
 
-RoaringSetBitForwardIterator Roaring::end() const {
+inline RoaringSetBitForwardIterator Roaring::end() const {
       return RoaringSetBitForwardIterator(*this, true);
 }
 
--- a/include/roaring/containers/run.h
+++ b/include/roaring/containers/run.h
@@ -74,7 +74,7 @@ void *run_container_deserialize(const char *buf, size_t buf_len);
 /*
  * Effectively deletes the value at index index, repacking data.
  */
-static void recoverRoomAtIndex(run_container_t *run, uint16_t index) {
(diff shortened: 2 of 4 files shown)


include/roaring/containers/run.h
```diff
@@ -74,7 +74,7 @@ void *run_container_deserialize(const char *buf, size_t buf_len);
 /*
  * Effectively deletes the value at index index, repacking data.
  */
-static void recoverRoomAtIndex(run_container_t *run, uint16_t index) {
+static inline void recoverRoomAtIndex(run_container_t *run, uint16_t index) {
```
lemire: Why would the "inline" keyword be needed here?

cpp/roaring.hh
```diff
@@ -506,11 +506,11 @@ public:
 };
 
 
-RoaringSetBitForwardIterator Roaring::begin() const {
+inline RoaringSetBitForwardIterator Roaring::begin() const {
```
lemire: why would inline be needed here?

cpp/roaring.hh
```diff
@@ -506,11 +506,11 @@ public:
...
       return RoaringSetBitForwardIterator(*this);
 }
 
-RoaringSetBitForwardIterator Roaring::end() const {
+inline RoaringSetBitForwardIterator Roaring::end() const {
```
lemire: why would inline be needed here?

lemire:
The change to the amalgamation script makes sense but I don't understand the 4 "inline" keywords you have added. If there are performance considerations, can you elaborate?

lemire:
I'll merge your PR because I think it probably does more good than harm, but I'd still like to understand why you put the "inline" keywords.

ksajme:
Sure, the "inline" keyword in current C++ does not instruct the compiler to perform an optimization: http://en.cppreference.com/w/cpp/language/inline . Confusingly, it allows multiple definitions as long as each is in a different compilation unit. This change is necessary to include roaring.hh multiple times in different C++ source files within the same program, for example. Without the fix there were linker errors including Roaring in large projects.

lemire:
@ksajme 

Good point. I wonder how we could test this out so such problems do not reappear later?

PR #75, opened 2017-01-25
Added 64-bit C++ version using std::map for the high-order bits

ksajme:
a 64-bit C++ implementation using the TreeMap methodology described in the http://r-libre.teluq.ca/930/1/Roaring64bits.pdf paper.

### DIFF
--- a/cpp/roaring64map.hh
+++ b/cpp/roaring64map.hh
@@ -1,0 +1,796 @@
+/*
+A C++ header for 64-bit Roaring Bitmaps, implemented by way of a map of many 32-bit Roaring Bitmaps.
+*/
+#ifndef INCLUDE_ROARING_64_MAP_HH_
+#define INCLUDE_ROARING_64_MAP_HH_
+
+#include <cstdarg>
+#include <cstdio>
+#include <algorithm>
+#include <new>
+#include <stdexcept>
+#include <utility>
+#include <map>
+#include <numeric>
+#include <limits>
+
+#include <roaring.hh>
+
+class Roaring64MapSetBitForwardIterator;
+
+class Roaring64Map{
+   public:
+    /**
+     * Create an empty bitmap
+     */
+    Roaring64Map() = default;
+
+    /**
+     * Construct a bitmap from a list of 32-bit integer values.
+     */
+    Roaring64Map(size_t n, const uint32_t *data) {
+        addMany(n, data);
+    }
+
+    /**
+     * Construct a bitmap from a list of 64-bit integer values.
+     */
+    Roaring64Map(size_t n, const uint64_t *data) {
+        addMany(n, data);
+    }
+
+    /**
+     * Copy constructor
+     */
+    Roaring64Map(const Roaring64Map &r) : roarings(r.roarings) { }
+
+    /**
+     * Construct a 64-bit map from a 32-bit one
+     */
+    Roaring64Map(const Roaring &r) { roarings.emplace(std::make_pair(0, r)); }
+
+    /**
+     * Construct a roaring object from the C struct.
+     *
+     * Passing a NULL point is unsafe.
+     */
+    Roaring64Map(roaring_bitmap_t *s) { roarings.emplace(std::make_pair(0, s)); }
+
+    /**
+     * Construct a bitmap from a list of integer values.
+     */
+    static Roaring64Map bitmapOf(size_t n...) {
+        Roaring64Map ans;
+        va_list vl;
+        va_start(vl, n);
+        for (size_t i = 0; i < n; i++) {
+            ans.add(va_arg(vl, uint64_t));
+        }
+        va_end(vl);
+        return ans;
+    }
+
+
+    /**
+     * Add value x
+     *
+     */
+    void add(uint32_t x) {
+        roarings[0].add(x);
+        roarings[0].setCopyOnWrite(copyOnWrite);
+    }
+    void add(uint64_t x) {
+        roarings[highBytes(x)].add(lowBytes(x));
+        roarings[highBytes(x)].setCopyOnWrite(copyOnWrite);
+    }
+
+    /**
+     * Add value n_args from pointer vals
+     *
+     */
+    void addMany(size_t n_args, const uint32_t *vals) {
+        for (size_t lcv = 0; lcv < n_args; lcv++) {
+            roarings[0].add(vals[lcv]);
+            roarings[0].setCopyOnWrite(copyOnWrite);
+        }
+    }
+    void addMany(size_t n_args, const uint64_t *vals) {
+        for (size_t lcv = 0; lcv < n_args; lcv++) {
+            roarings[highBytes(vals[lcv])].add(lowBytes(vals[lcv]));
+            roarings[highBytes(vals[lcv])].setCopyOnWrite(copyOnWrite);
+        }
+    }
+
+    /**
+     * Remove value x
+     *
+     */
+    void remove(uint32_t x) { roarings[0].remove(x); }
+    void remove(uint64_t x) {
+        auto roaring_iter = roarings.find(highBytes(x));
+        if (roaring_iter != roarings.cend())
+            roaring_iter->second.remove(lowBytes(x));
+    }
+
+    /**
+     * Return the largest value (if not empty)
+     *
+     */
+    uint64_t maximum() const {
+        for (auto roaring_iter = roarings.crbegin(); roaring_iter != roarings.crend(); ++roaring_iter) {
+            if (!roaring_iter->second.isEmpty()) {
+                return (uint64_t(roaring_iter->first) << 32) | uint64_t(roaring_iter->second.maximum());
+            }
+        }
+        return std::numeric_limits<uint64_t>::min();
+    }
+
+
+    /**
+     * Return the smallest value (if not empty)
+     *
+     */
+    uint64_t minimum() const {
+        for (auto roaring_iter = roarings.cbegin(); roaring_iter != roarings.cend(); ++roaring_iter) {
+            if (!roaring_iter->second.isEmpty()) {
+                return (uint64_t(roaring_iter->first) << 32) | uint64_t(roaring_iter->second.minimum());
+            }
+        }
+        return std::numeric_limits<uint64_t>::max();
+    }
+
+
+    /**
+     * Check if value x is present
+     */
+    bool contains(uint32_t x) const {
+        return roarings.at(0).contains(x);
+    }
+    bool contains(uint64_t x) const {
+        return roarings.at(highBytes(x)).contains(lowBytes(x));
+    }
+
+    /**
+     * Destructor
+     */
+    ~Roaring64Map() = default;
+
+    /**
+     * Copies the content of the provided bitmap, and
+     * discards the current content.
+     */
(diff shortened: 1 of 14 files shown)


lemire:
This looks good to me. Clean code. Complete with unit tests.

Maybe @fsaintjacques and @madscientist want to review before we pull this into master?

lemire:
c.c. @owenkaser

lemire:
@fsaintjacques 

(Thanks for the thorough review.)

I am not eager for us to use non-standard types. For example, I'd like for this C++ API to work more or less unchanged under Visual Studio or other compilers. And who wants to deal with non-standard types on a regular fashion? You can't even print these things easily! I think we should favor "standard types" as much as possible.

The problem you identify is real, however. I am not dismissing your concern in the least.

lemire:
@fsaintjacques
@ksajme 


We really, really don't want to make the API hard to use with custom data types where we can avoid it. The case where the cardinality is 1<<64 is unique and can be optimized out at the application level. We don't want everyone to suffer because of this one case.

The idea that we should be able to hold data structures as large as memory would allow is quickly invalidated... The following program will crash as a 32-bit binary when sizeof(size_t) is 4...

```
#include <iostream>
#include <vector>


int main() {
  std::cout << sizeof(size_t) << std::endl;
  std::vector<bool> x;
  for(uint64_t z = 0 ; z < UINT64_C(1) << 32; ++z)
    x.push_back(0);
  std::cout << x.size() << std::endl;
}
```

Try it...
```bash
g++ -m32 -O3 -o toobig toobig.cpp
./toobig
```

You should get ``std::length_error``.


The corresponding thing should happen on a 64-bit machine with a much larger data structure but, of obvious reasons I cannot test it.

So we can add a test of the type "if computed cardinality is zero and the roaring bitmap is not empty, then throw an exception".

It is not perfect but it is as good as what STL does.

*And* we should document it, of course.

Thoughts?

lemire:
@ksajme

So unless @fsaintjacques  comes back with something better, I recommend you modify your API so that when the cardinality overflows (it is equal to zero but the bitmap is not empty) then you throw an exception.

madscientist:
Great stuff @ksajme.  Regarding the cardinality issue @fsaintjacques; I see your point about that.  However, while it might be possible to create a completely full roaring32 in a reasonable size using RLE, this implementation uses a std::map to hold the individual 32bit roaring structures without any RLE etc. in the upper 32bits.  I don't know how large a full roaring32 is, but an empty one is 44 bytes and I doubt it's smaller than that.  So, just the memory needed to store 2^32 full roaring32 must be at least 176G (4G * 44), and that's not counting whatever overhead is required for that many std::map elements.

A different implementation could definitely reduce this significantly.  I don't know if it's worthwhile to force the API on this implementation, given that it requires non-standard types, to avoid changing it if/when a new fancier implementation comes along.

I wrote some ideas for how to resolve this but @lemire already mentioned them... the only other one would be to return a structure instead of a uint64_t which contains a cardinality value and an extra boolean that means "full", but given all the above I have a hard time justifying that.

ksajme:
@fsaintjacques thank you for the feedback. I'll address the issues that you found. Interesting off-by-one problem.

Visual Studio doesn't have the 128 bit numbers we need but you came up with a good solution @lemire . I'll implement it.

I forgot to add the 64-bit header to the amalgamation and the demo. I'll go ahead and do that.

lemire:
@madscientist 

Agreed.

With run compression, a full 32-bit roaring might easily fit in 64 bytes. So we are talking about maybe 256 GB to create a full 64-bit roaring bitmap using the approach indicated here. So, yes, it is huge. That would not be the right way to go about it. You don't want to be using hundreds of gigabytes for something like this.

madscientist:
@lemire yes... I think the first person who needs something like that should be drafted into enhancing this implementation :-).  This implementation will run into problems in situations where you set bits whose values are scattered across a significant portion of the upper 32bit space.  An implementation that handles that more gracefully would be very cool, no doubt... but an enormous amount of effort I'd imagine.

fsaintjacques:
Exception might be overkill, I'd prefer a struct with a bool that says full. Much less intrusive.

tests/cpp_unit.cpp
```diff
@@ -214,6 +220,106 @@ void test_example_cpp(bool copy_on_write) {
...
+    }
+    for (uint64_t i = 14000000000000000100ull; i < 14000000000000001000ull; i++) {
+        r1.add(i);
+    }
+
+    // check whether a value is contained
+    assert_true(r1.contains((uint64_t)14000000000000000500ull));
+
+    // compute how many bits there are:
+    uint32_t cardinality = r1.cardinality();
```
ksajme: it should

tests/cpp_unit.cpp
```diff
@@ -214,6 +220,106 @@ void test_example_cpp(bool copy_on_write) {
...
+    std::cout << "Cardinality = " << cardinality << std::endl;
+
+    // if your bitmaps have long runs, you can compress them by calling
+    // run_optimize
+    uint32_t size = r1.getSizeInBytes();
```
ksajme: fixed this part of the test as well

madscientist:
The advantage of an exception (to me) is that for people who are confident their machine will keel over long, long before they have to worry about it (which is most of us) they can just ignore the exception, or catch it in main() or whatever for sanity.  Personally I'd prefer to deal with an exception in the vanishingly unlikely case that you really do have a completely full roaring64, than pay the price for returning a 12+byte structure on every cardinality check and the added complexity of dealing with a returned structure and checking two different values.

Plus as @lemire points out this behavior matches other parts of the STL.

Another option would be to provide an additional non-throwing cardinality method, for those who felt it might be desirable.  In that case they'd need to check "isempty" if the cardinality was 0, to see whether it was really 0 bits set or if it was 2^64 bits set.  Or something like that.  Of course this can be easily wrapped by the user themselves if they need it.

Anyway that's my $0.02.  I'll leave it up to you all to sort it.

One note: if you do throw would it be worthwhile to throw a roaring-specific subclass of std::length_error instead of the standard exception itself... just to make it simpler to catch?  Not sure, just a consideration.

tests/cpp_unit.cpp
```diff
@@ -214,6 +220,106 @@ void test_example_cpp(bool copy_on_write) {
...
+    char *serializedbytes = new char[expectedsize];
+    r1.write(serializedbytes);
+    Roaring64Map t = Roaring64Map::read(serializedbytes);
+    assert_true(r1 == t);
+    delete[] serializedbytes;
+
+    // we can iterate over all values using custom functions
+    uint64_t counter = 0;
+    r1.iterate(roaring_iterator_sumall64, &counter);
+    /**
```
ksajme: you'd like me to remove the comment? it's emulating the one in the 32-bit test above

cpp/roaring64map.hh
```diff
@@ -0,0 +1,796 @@
...
+                if (!map_entry.second.isSubset(roaring_iter->second))
+                    return false;
+        }
+        return true;
+    }
+
+    /**
+    * Returns true if the bitmap is strict subset of the other.
+    */
+    bool isStrictSubset(const Roaring64Map &r) const { return isSubset(r) && !(*this == r); }
```
ksajme: good point

cpp/roaring64map.hh
```diff
@@ -0,0 +1,796 @@
...
+                return previous;
+            });
+    }
+
+    /**
+     * Return true if the two bitmaps contain the same elements.
+     */
+    bool operator==(const Roaring64Map &r) const {
+        // we cannot use operator == on the map because either side may contain empty Roaring Bitmaps
+        auto lhs_iter = roarings.cbegin();
```
ksajme: I added a number of comments to improve readability. The maps do not remove empty roaring elements unless shrinkToFit() is called. Thus we need a more complicated equality that skips over any empties found.

cpp/roaring64map.hh
```diff
@@ -0,0 +1,796 @@
...
+     *  all the values with ptr (can be NULL) as the second parameter of each
+     * call.
+     *
+     *  roaring_iterator64 is simply a pointer to a function that returns void,
+     *  and takes (uint64_t,void*) as inputs.
+     */
+    void iterate(roaring_iterator64 iterator, void *ptr) const {
+        std::for_each(roarings.begin(), roarings.cend(),
+            [=](const std::pair<uint32_t, Roaring>& map_entry) {
+                roaring_iterate64(map_entry.second.roaring, iterator, uint64_t(map_entry.first) << 32, ptr);
```
ksajme: I added an inline function that stitches two 32-bit values back into a 64-bit one

cpp/roaring64map.hh
```diff
@@ -0,0 +1,796 @@
...
+
+    /**
+     * If the size of the roaring bitmap is strictly greater than rank, then
+     this
+       function returns true and set element to the element of given rank.
+       Otherwise, it returns false.
+     */
+    bool select(uint64_t rank, uint32_t *element) const {
+        for (const auto& map_entry : roarings) {
+            uint64_t sub_cardinality = (uint32_t)map_entry.second.cardinality();
```
ksajme: this was supposed to be a (uint64_t) cast to keep Visual Studio from complaining

ksajme:
I'll throw an exception then. A full data structure will be an exceptional circumstance due to the current limitations of physical memory. Meanwhile the common case will be unaffected with an unusual return type.

lemire:
@ksajme 

Could you add a ``isFull()`` method? In this manner, the user could first check whether the roaring is full... and if so then he/she knows that the cardinality method will not return the right value. The call to ``isFull()`` could be dirt cheap in practice because it could start by checking the size of the map. In all practical cases, the map would be too small to generate a full bitmap.

So one could write exception-free code like so...

```
uint128_t card = b.isFull() ?  1<< 64 : b.cardinality()
```

lemire:
@ksajme  Please add yourself to the AUTHORS file.

lemire:
@ksajme 

Seems that there are compilation errors having to do with ``emplace``: 

https://travis-ci.org/RoaringBitmap/CRoaring/builds/195387727

We probably can solve this with something like...

```
#ifdef __cplusplus >= 201103L
// do something wild
#else
// do boring stuff
#endif
```

ksajme:
You're right, I think this is:
https://gcc.gnu.org/bugzilla/show_bug.cgi?id=44436

Which is a lack of emplace for maps in gcc's C++11 libstdc++ prior to version 4.8. Although Travis is using Clang, I'm guess it's building against the gcc STL.

I wrote a macro to detect old versions of the gcc STL and use insert instead. Let's see if Travis agrees with me...

lemire:
Ah. Ah. Tests are passing now.

lemire:
@fsaintjacques I trust that you feel that it is acceptable now?

Ezibenroc:
I am wondering, would it be possible (and easy) to use this C++ code in the C library? Or would it be better to reimplement it in C from scratch?

lemire:
@Ezibenroc I think it would be a lot easier to code a 64-bit data structure in C now that we have a fully worked out example in C. All one needs is an ordered int-to-pointer map in C. I am sure we can find a decent and lightweight C implementation we can adapt.

fsaintjacques:
Go ahead, thanks for the contribution @ksajme .

lemire:
❤️

PR #77, opened 2017-02-21
Embedded C object inside C++ object

ksajme:
changed the pointer to a CRoaring object inside of the C++ class into the object itself,
avoiding a malloc and free in the constructor/destructor,
this makes little difference to the 32-bit C++ class but speeds up adds to the 64-bit class by a few percent,
also did a few minor things

Bug fix:
    C++ iterators now stop at max uint32_t instead of comparing equal to the end iterator at this point
Performance:
   removed virtual keyword for C++ iterator class's destructor since it will cause most compilers
   to generate a vtable for each object, bloating it up
   (the lack of the keyword is safe in this case unless a user subclasses the iterator
   and that subclass is deleted through a pointer with the type of the base class)
Feature:
    added a toString member function to the C++ classes which makes STL strings

### DIFF
--- a/cpp/roaring.hh
+++ b/cpp/roaring.hh
@@ -9,6 +9,7 @@ A C++ header for Roaring Bitmaps.
 #include <algorithm>
 #include <new>
 #include <stdexcept>
+#include <string>
 #include <roaring/roaring.h>
 
 class RoaringSetBitForwardIterator;
@@ -18,38 +19,45 @@ class Roaring {
     /**
      * Create an empty bitmap
      */
-    Roaring() : roaring(NULL) {
-        roaring = roaring_bitmap_create();
-        if (roaring == NULL) {
+    Roaring() {
+        bool is_ok = ra_init(&roaring.high_low_container);
+        if (!is_ok) {
             throw std::runtime_error("failed memory alloc in constructor");
         }
+        roaring.copy_on_write = false;
     }
 
     /**
      * Construct a bitmap from a list of integer values.
      */
-    Roaring(size_t n, const uint32_t *data) {
-        roaring = roaring_bitmap_of_ptr(n, data);
-        if (roaring == NULL) {
-            throw std::runtime_error("failed memory alloc in constructor");
-        }
+    Roaring(size_t n, const uint32_t *data) : Roaring() {
+        roaring_bitmap_add_many(&roaring, n, data);
     }
     /**
      * Copy constructor
      */
-    Roaring(const Roaring &r) : roaring(NULL) {
-        roaring = roaring_bitmap_copy(r.roaring);
-        if (roaring == NULL) {
+    Roaring(const Roaring &r) {
+        bool is_ok = ra_copy(&r.roaring.high_low_container, &roaring.high_low_container,
+                             r.roaring.copy_on_write);
+        if (!is_ok) {
             throw std::runtime_error("failed memory alloc in constructor");
         }
+        roaring.copy_on_write = r.roaring.copy_on_write;
     }
 
     /**
      * Construct a roaring object from the C struct.
      *
      * Passing a NULL point is unsafe.
+     * the pointer to the C struct will be invalid after the call.
      */
-    Roaring(roaring_bitmap_t *s) : roaring(s) {}
+    Roaring(roaring_bitmap_t *s) {
+        // steal the interior struct
+        roaring.high_low_container = s->high_low_container;
+        roaring.copy_on_write = s->copy_on_write;
+        // deallocate the old container
+        free(s);
+    }
 
     /**
      * Construct a bitmap from a list of integer values.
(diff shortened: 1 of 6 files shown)


cpp/roaring.hh
```diff
@@ -18,38 +19,45 @@ class Roaring {
...
 
     /**
      * Construct a roaring object from the C struct.
      *
      * Passing a NULL point is unsafe.
-     */
-    Roaring(roaring_bitmap_t *s) : roaring(s) {}
+     * the pointer to the C struct will be invalid after the call.
+     */
+    Roaring(roaring_bitmap_t *s) {
```
ksajme: This is an interesting case. It isn't possible to replicate exactly the old behavior of this function with an embedded roaring_bitmap_t because we have to use new memory in a constructor. I've selected the approach because it seems the least disruptive: There isn't a copy so it hasn't changed in terms of semantics or much in terms of performance. It frees the memory it is given so the contract hasn't changed, all of the memory is still now handled by the C++ object. The only downside is that the given pointer is no longer a valid alias to the data structure.

cpp/roaring.hh
```diff
@@ -401,13 +445,14 @@ class Roaring {
...
         }
-        for (size_t k = 0; k < n; ++k) x[k] = inputs[k]->roaring;
+        for (size_t k = 0; k < n; ++k) x[k] = &inputs[k]->roaring;
 
-        Roaring ans(NULL);
-        ans.roaring = roaring_bitmap_or_many(n, x);
-        if (ans.roaring == NULL) {
+        roaring_bitmap_t* c_ans = roaring_bitmap_or_many(n, x);
+        if (c_ans == NULL) {
+            free(x);
```
ksajme: the lack of a free here was a leak on the exceptional case

cpp/roaring.hh
```diff
@@ -462,18 +502,26 @@ public:
   }
 
   bool operator<(const type_of_iterator &o) {
+    if (!i.has_value) return false;
+    if (!o.i.has_value) return true;
```
ksajme: not checking has_value caused the end of the iterator to compare equal to the last valid value of the iterator for every function

cpp/roaring.hh
```diff
@@ -489,26 +537,28 @@ public:
...
         i.container_index = INT32_MAX;
         i.has_value = false;
         i.current_value = UINT32_MAX;
     } else {
-      roaring_init_iterator(parent.roaring, &i);
+      roaring_init_iterator(&parent.roaring, &i);
     }
   }
 
-  virtual ~RoaringSetBitForwardIterator() {
```
ksajme: Virtual here only helps the case where a user subclasses this object and uses it in a polymorphic way (including deleting it from a polymorphic context). I would argue that it isn't the job of this library to worry about this case and that it is very unlikely that a user would want to do that regardless.

include/roaring/roaring_array.h
```diff
@@ -1,5 +1,8 @@
 #ifndef INCLUDE_ROARING_ARRAY_H
 #define INCLUDE_ROARING_ARRAY_H
+#ifdef __cplusplus
+extern "C" {
+#endif
```
ksajme: I needed to add another extern "C" block since this file is referred to from within the C++ code

cpp/roaring64map.hh
```diff
@@ -646,6 +647,44 @@ class Roaring64Map{
     }
 
     /**
+     * Print the content of the bitmap into a string
+     */
+    std::string toString() const {
```
ksajme: I decided to use an STL string for output. C++ users nowadays are either using the STL or a custom string class that we cannot guess at.

cpp/roaring.hh
```diff
@@ -300,15 +310,27 @@ class Roaring {
...
+                buf += 1;
+            } else
+                throw std::runtime_error("bad serialization type designator in read input");
+        } {
+            Roaring ans;
+            bool is_ok = ra_portable_deserialize(&ans.roaring.high_low_container, buf);
+            if(!is_ok) {
+                throw std::runtime_error("failed memory alloc while reading");
+            }
+            return ans;
```
ksajme: this function is a bit less elegant and duplicates some of the code from the C source files but I think it is a fair trade for not perturbing the C at all

include/roaring/roaring.h
```diff
@@ -168,7 +168,6 @@ void roaring_bitmap_andnot_inplace(roaring_bitmap_t *x1,
  */
 void roaring_bitmap_free(roaring_bitmap_t *r);
 
-
```
ksajme: apparently I felt the need to move this empty line to roaring.c for some reason! oops

ksajme:
Great job on getting rid of the allocation and pointer in the iterator @lemire .

I had the idea to do the same thing for the 32-bit C++ object.

lemire:
> removed virtual keyword for C++ iterator class's destructor since it will cause most compilers to generate a vtable for each object, bloating it up

Should it be declared ``final``?

http://en.cppreference.com/w/cpp/language/final

ksajme:
I considered that. And I also considered adding final to every class in the library. However it is safe to inherit from them if you don't polymorphically destroy them. For the user though, it isn't hard to take the keyword off. We could add final as a warning to this class or to all of them or to none of them.

lemire:
Seems like a minor point anyhow. I do not expect anyone to extend the iterators.

ksajme:
I'll go ahead and put final on both of them.

lemire:
LGTM

lemire:
This is part of the latest release. I have also significantly improved support for Visual Studio, fixing one significant bug in the "maximum" function (bug specific to Visual Studio). I have also improved the documentation pertaining to Visual Studio.

ksajme:
I saw your changes related to Visual Studio and they look great. I haven't had a chance to build on Windows since then but have no doubt it will be a smoother experience.

lemire:
@ksajme 

Actually, I'd love for someone to go over my instructions and maybe improve them...
https://github.com/RoaringBitmap/CRoaring#building-visual-studio-under-windows

It should work "out of the box" without tinkering.

ksajme:
Well done, it works exactly as described. I didn't expect it to be so seamless so quickly.

I also tested it with the Visual Studio 2017 release candidate so we should be set for awhile.

Better remember to add VisualStudio or whatever to .gitignore or you might get someone's Visual Studio build directory in a pull request...

lemire:
@ksajme Excellent.

I read that Visual Studio 2017 release candidate supports ``cmake`` directly. Did you try that as well? (I have not touched Visual Studio 2017.)

ksajme:
I just tried it, unfortunately it does not work. It fails with the line:
C:\...\build\Visual Studio 15 x64\tests\LINK : fatal error LNK1104: cannot open file '..\src\Debug\roaring.lib'

This might be related to the fact that the blog:
https://blogs.msdn.microsoft.com/vcblog/2016/10/05/cmake-support-in-visual-studio/
shows four options in the drop down (see the first picture under "Configuring CMake"

Roaring shows only two:
Visual Studio 15 x86
Visual Studio 15 x64
don't forget that: (Visual Studio 2015 == Visual Studio 14 && Visual Studio 2017 == Visual Studio 15)

So at some point it might forget the fact that it's in Release mode and default to looking in a Debug directory.

lemire:
@ksajme 

Did you check what is in '..\src\Debug'? My guess is that you will find a dynamic library ('..\src\Debug\roaring.dll'). That's the default right now in our ``cmake`` configuration.  I think that to get '..\src\Debug\roaring.lib', we need to invoke ``cmake`` with ``BUILD_STATIC=ON``.  Any chance you could try it somehow?

Alternatively, we could investigate the issue of generating a dynamic library with Visual Studio. I will open an issue.

ksajme:
Your guess sounds better than mine. I'll confirm in a bit.

lemire:
I'll change the ``cmake`` configuration so that under Visual Studio, we default to a static library.

lemire:
Try it again with this change in our default: https://github.com/RoaringBitmap/CRoaring/commit/80e5aa4e8b6b6d41af2ab718dfc546390d40cd61

ksajme:
This still works with 2015 and 2017 the manual way, but without BUILD_STATIC=ON so that's good.

For the built-in GUI method, it is getting farther but choking on:
C:\...\croaring\benchmarks\numbersfromtextfiles.h(6): fatal error C1083: Cannot open include file: 'dirent.h': No such file or directory

Obviously it shouldn't be poking around there. Let me see if I can figure out why.

lemire:
@ksajme 

Right, as you understand, it should not try to build anything that depends on ``dirent.h``.

ksajme:
Working now. I opened a pull request: https://github.com/RoaringBitmap/CRoaring/pull/79

PR #79, opened 2017-02-23
Fixed IDE compilation on Visual Studio 2017 and added README instructions

ksajme:


benchmarks/CMakeLists.txt
```diff
@@ -1,11 +1,10 @@
...
+# We exclude POSIX tests from Visual Studio default build
+if(NOT WIN32)
+    add_c_benchmark(real_bitmaps_benchmark)
+endif()
 add_c_benchmark(bitset_container_benchmark)
 add_c_benchmark(array_container_benchmark)
 add_c_benchmark(run_container_benchmark)
-
-# We exclude POSIX tests from Visual Studio default build 
-set_target_properties(real_bitmaps_benchmark PROPERTIES EXCLUDE_FROM_DEFAULT_BUILD 1)
```
ksajme: EXCLUDE_FROM_DEFAULT_BUILD is ignored by Visual Studio 2017's CMake components.

benchmarks/CMakeLists.txt
```diff
@@ -1,11 +1,10 @@
 set (BENCHMARK_DATA_DIR "${CMAKE_CURRENT_SOURCE_DIR}/realdata/") # does this ever get used?
 
 add_c_benchmark(create_benchmark)
-add_c_benchmark(real_bitmaps_benchmark)
+# We exclude POSIX tests from Visual Studio default build
+if(NOT WIN32)
```
ksajme: Future work is to check for a POSIX runtime rather than use this coarse switch. But it doesn't seem to be doable from CMake. Excluding this one test isn't a big deal in the case of an odd Win32 POSIX-compatible system.

lemire:
Looks excellent.

PR #80, opened 2017-02-28
made Visual Studio's warnings quieter the hard way

ksajme:
"Roaring" should not refer to the amount of warning text generated. This fixes a number of the warnings about code style (and one real bug) which are generated by Visual Studio on the default warning level. Note that this isn't quite all of them, only those in the path of my application.

### DIFF
--- a/cpp/roaring64map.hh
+++ b/cpp/roaring64map.hh
@@ -248,7 +248,7 @@ class Roaring64Map{
             throw std::length_error("bitmap is full, cardinality is 2^64, "
                                     "unable to represent in a 64-bit integer");
         }
-        return std::accumulate(roarings.cbegin(), roarings.cend(), 0,
+        return std::accumulate(roarings.cbegin(), roarings.cend(), (uint64_t)0,
             [](uint64_t previous, const std::pair<uint32_t, Roaring>& map_entry) {
                 return previous + map_entry.second.cardinality();
             });
@@ -461,7 +461,7 @@ class Roaring64Map{
             if (rank < sub_cardinality) {
                 *element = ((uint64_t)map_entry.first) << 32;
                 // assuming little endian
-                return map_entry.second.select(rank, ((uint32_t*)element));
+                return map_entry.second.select((uint32_t)rank, ((uint32_t*)element));
             }
             rank -= sub_cardinality;
         }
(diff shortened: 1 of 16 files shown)


cpp/roaring64map.hh
```diff
@@ -501,8 +501,8 @@ class Roaring64Map{
     size_t write(char *buf, bool portable = true) const {
         const char* orig = buf;
         // push map size
-        *((uint32_t*)buf) = roarings.size();
-        buf += sizeof(uint32_t);
+        *((uint64_t*)buf) = roarings.size();
```
ksajme: an unfortunate mistake, carnality being a 64-bit number for a 32-bit data structure strikes again!

include/roaring/bitset_util.h
```diff
@@ -16,12 +16,12 @@ static inline void bitset_set_range(uint64_t *bitmap, uint32_t start,
     uint32_t endword = (end - 1) / 64;
     if (firstword == endword) {
         bitmap[firstword] |= ((~UINT64_C(0)) << (start % 64)) &
-                             ((~UINT64_C(0)) >> ((-end) % 64));
+                             ((~UINT64_C(0)) >> ((~end + 1) % 64));
```
ksajme: MSVC doesn't like applying unary minus to unsigned numbers:
warning C4146: unary minus operator applied to unsigned type, result still unsigned

include/roaring/containers/bitset.h
```diff
@@ -154,9 +154,9 @@ static inline bool bitset_container_add(bitset_container_t *bitset,
     const int index = pos & 63;
     const uint64_t new_word = old_word | (UINT64_C(1) << index);
     const uint64_t increment = (old_word ^ new_word) >> index;
-    bitset->cardinality += increment;
+    bitset->cardinality += (uint32_t)increment;
     bitset->array[pos >> 6] = new_word;
-    return increment;  // 0 == false, 1 == true
+    return increment > 0;
```
ksajme: warning C4800: 'const uint64_t': forcing value to bool 'true' or 'false' (performance warning)
this is probably easier to read too

src/containers/convert.c
```diff
@@ -267,11 +267,11 @@ void *convert_run_optimize(void *c, uint8_t typecode_original,
             uint64_t cur_word_with_1s = cur_word | (cur_word - 1);
 
             int run_end = 0;
-            while (cur_word_with_1s == UINT64_C(-1) &&
+            while (cur_word_with_1s == UINT64_C(0xFFFFFFFFFFFFFFFF) &&
```
ksajme: I settled on a long string of 'F's

lemire:
Excellent.

PR #82, opened 2017-03-10
Fixed some bugs in the 64-bit layer

ksajme:
A few bugs discovered when making heavy use of the 64-bit containers.

### DIFF
--- a/cpp/roaring64map.hh
+++ b/cpp/roaring64map.hh
@@ -173,11 +173,11 @@ class Roaring64Map{
      * modified.
      */
     Roaring64Map &operator&=(const Roaring64Map &r) {
-        for (const auto& map_entry : r.roarings) {
-            if (roarings.count(map_entry.first) == 1) {
-                roarings[map_entry.first] &= map_entry.second;
-                roarings[map_entry.first].setCopyOnWrite(copyOnWrite);
-            }
+        for (auto& map_entry : roarings) {
+            if (r.roarings.count(map_entry.first) == 1)
+                map_entry.second &= r.roarings.at(map_entry.first);
+            else
+                map_entry.second = Roaring();
         }
         return *this;
     }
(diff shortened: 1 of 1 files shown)


cpp/roaring64map.hh
```diff
@@ -173,11 +173,11 @@ class Roaring64Map{
      * modified.
      */
     Roaring64Map &operator&=(const Roaring64Map &r) {
-        for (const auto& map_entry : r.roarings) {
-            if (roarings.count(map_entry.first) == 1) {
-                roarings[map_entry.first] &= map_entry.second;
-                roarings[map_entry.first].setCopyOnWrite(copyOnWrite);
-            }
```
ksajme: For the and operation: the left hand side map was erroneously holding on to Roarings representing ranges that weren't present in the right hand side in the previous implementation.

cpp/roaring64map.hh
```diff
@@ -826,7 +826,7 @@ public:
     Roaring64MapSetBitForwardIterator(const Roaring64Map& parent, bool exhausted = false)
         : map_end(parent.roarings.cend())
     {
-        if(exhausted) {
+        if(exhausted || parent.roarings.cbegin() == parent.roarings.cend()) {
```
ksajme: An empty map needs special handling to create an exhausted iterator or this constructor will attempt to extract the first Roaring.

cpp/roaring64map.hh
```diff
@@ -842,10 +842,7 @@ public:
 
     ~Roaring64MapSetBitForwardIterator() = default;
 
-    Roaring64MapSetBitForwardIterator(
-        const Roaring64MapSetBitForwardIterator &o)
-        : map_iter(o.map_iter), map_end(o.map_end), i(o.i) {
-    }
+    Roaring64MapSetBitForwardIterator(const Roaring64MapSetBitForwardIterator &o) = default;
```
ksajme: This is the default copy constructor so might as well not add the details.

cpp/roaring64map.hh
```diff
@@ -826,7 +826,7 @@ public:
     Roaring64MapSetBitForwardIterator(const Roaring64Map& parent, bool exhausted = false)
         : map_end(parent.roarings.cend())
     {
-        if(exhausted) {
+        if(exhausted || parent.roarings.empty()) {
```
ksajme: An empty map needs special handling to create an exhausted iterator or this constructor will attempt to extract the first Roaring.

lemire:
LGTM.

PR #85, opened 2017-03-14
addressed getSizeInBytes() bug and made serialized 64-bit maps simpler

ksajme:
With this update I'm changing to an on-disk format for 64-bit maps that doesn't store the size of the interior Roarings.

This is what I wanted to do originally by modifying the C API to return the amount of bytes read but I noticed I could achieve the same thing by using getSizeInBytes() after retrieval of each Roaring.

Note that although this breaks compatibility with old serialized maps, I think it's less of an issue since the getSizeInBytes() bug made it difficult to serialize in the first place!

Also if the C API was ever changed to return the bytes read, there would be no further need to modify the on-disk format. I think it is here to stay.

### DIFF
--- a/cpp/roaring64map.hh
+++ b/cpp/roaring64map.hh
@@ -508,12 +508,8 @@ class Roaring64Map{
                 // push map key
                 *((uint32_t*)buf) = map_entry.first;
                 buf += sizeof(uint32_t);
-                // byte count in Roaring goes here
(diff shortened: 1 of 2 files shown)


lemire:
I am fine with this since we never promised that the 64-bit implementation would have a stable serialized format.

lemire:
Let us make this part of the next release.

PR #88, opened 2017-03-26
Fix roaring_bitmap_intersect.

Ezibenroc:
There was a bug when testing the intersection between a bitset and a run container with only one run.

For example, this test was failing:
```c
void test_intersect_small_run_bitset() {
    roaring_bitmap_t *rb1 = roaring_bitmap_from_range(0, 1, 1);
    roaring_bitmap_t *rb2 = roaring_bitmap_from_range(1, 8194, 2);
    assert_false(roaring_bitmap_intersect(rb1, rb2));
    roaring_bitmap_free(rb1);
    roaring_bitmap_free(rb2);
}
```

I did not add this piece of code in the test suite of CRoaring, because more extensive tests may be better.


lemire:
Thanks. I'll be adding your test.

PR #89, opened 2017-04-12
Added move constructors

WojciechMula:
Since C++ lib is compiled with -std=c++11 flag, I suppose that using features introduced by C++11 is OK. :)

### DIFF
--- a/cpp/roaring.hh
+++ b/cpp/roaring.hh
@@ -509,8 +509,7 @@ class Roaring {
                         ((iter_data *)inner_iter_data)->first_char;
                     ((iter_data *)inner_iter_data)->str +=
                         std::to_string(value);
-                    if (((iter_data *)inner_iter_data)->first_char == '{')
-                        ((iter_data *)inner_iter_data)->first_char = ',';
+                    ((iter_data *)inner_iter_data)->first_char = ',';
                     return true;
                 },
                 (void *)&outer_iter_data);
--- a/cpp/roaring.hh
+++ b/cpp/roaring.hh
@@ -45,6 +45,18 @@ class Roaring {
         }
         roaring.copy_on_write = r.roaring.copy_on_write;
     }
+    /**
+     * Move constructor
+     */
+    Roaring(Roaring &&r) {
+        bool is_ok = ra_init(&roaring.high_low_container);
+        if (!is_ok) {
+            throw std::runtime_error("failed memory alloc in constructor");
+        }
+        roaring.copy_on_write = false;
+        
+        std::swap(roaring, r.roaring);
+    }
 
     /**
      * Construct a roaring object from the C struct.
--- a/cpp/roaring.hh
+++ b/cpp/roaring.hh
@@ -45,17 +45,19 @@ class Roaring {
         }
         roaring.copy_on_write = r.roaring.copy_on_write;
     }
+
     /**
-     * Move constructor
+     * Move constructor. The moved object remains valid, i.e.
+     * all methods can still be called on it.
      */
     Roaring(Roaring &&r) {
-        bool is_ok = ra_init(&roaring.high_low_container);
+        roaring = std::move(r.roaring);
+
+        // left the moved object in a valid state
+        bool is_ok = ra_init_with_capacity(&r.roaring.high_low_container, 1);
         if (!is_ok) {
             throw std::runtime_error("failed memory alloc in constructor");
         }
-        roaring.copy_on_write = false;
-        
-        std::swap(roaring, r.roaring);
     }
 
     /**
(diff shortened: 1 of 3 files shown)


lemire:
Wow. Yes, I suppose so.

lemire:
@ksajme Want to have a look?

lemire:
@WojciechMula  We'll merge this, but I'd like for someone else to independently review it. Let me see if I can ping someone interested...

lemire:
Maybe @fsaintjacques ?

cpp/roaring.hh
```diff
@@ -45,6 +45,18 @@ class Roaring {
...
+     * Move constructor
+     */
+    Roaring(Roaring &&r) {
+        bool is_ok = ra_init(&roaring.high_low_container);
+        if (!is_ok) {
+            throw std::runtime_error("failed memory alloc in constructor");
+        }
+        roaring.copy_on_write = false;
+        
+        std::swap(roaring, r.roaring);
```
gssiyankai: Is there a reason why we don't just do 
```
roaring = std::move(r.roaring)
```
here ?

lemire:
Ah. Thanks @gssiyankai

cpp/roaring.hh
WojciechMula: Move semantic requires to leave the right-side argument in a valid state. The desired behaviour is:

```
Roaring foo;
foo.add(1);

Roaring bar(std::move(foo)); // bar got data from foo, but
foo.add(2); // <<< this still must be possible
```

If we used merely std::move as you wrote, it would transfer all pointers from the underlaying roaring_bitmap_t structure. Transfer of bare pointers means setting them NULL on the right side. And then the last line in the above example would cause segfault.

lemire: > Transfer of bare pointers means setting them NULL on the right side.

Is this true? So if I type ``uint32_t* y = std::move(x);``, then ``x``will be NULL?

WojciechMula: @lemire No, I was wrong. However, the segfault occurs, I need to figure out why.

lemire: Does the destructor get called?

lemire:
@WojciechMula I'm 99% sure that your code is correct... but I'd feel better if we could have a well-reasoned answer to @gssiyankai 's comment.

ksajme:
Cool, while you're exploring move semantics will you also do the move assignment operator?

First to clear up what's legal for the state of the moved-from object here's what the STL requires of its own objects:
Unless otherwise specified, all standard library objects that have been moved from are placed in a valid but unspecified state. That is, only the functions without preconditions, such as the assignment operator, can be safely used on the object after it was moved from

Swapping the moved-to and the moved-from is an easy way to conform to the above but it may be more optimal to leave the moved-from object empty somehow. The moved-from object certainly needs to be destructable because the destructor will inevitably be called on it.

Let's examine the possibilities for Roaring (the 32-bit C++ object):

std::swap does something like this:

template<typename T> void swap(T& t1, T& t2) {
    T temp(std::move(t1));
    t1 = std::move(t2);
    t2 = std::move(temp);
} // (Note for the pedantic, T will get destructed here while holding a shared roaring but @WojciechMula is not seeing a segfault with swap. The actual implementation on the observed STL may be something equivalent but different)

So you'll get three moves.

= std::move will do a member-wise move of everything inside r.roaring directly into this->roaring. Note that this is actually the same as just omitting the std::move and doing member-wise copy since roaring_bitmap_t is a C struct.

The issue with std::move is that both Roarings will have roaring members which are discrete but identical. This means they will contain pointers to the same stuff and you will get a double free when they are deconstructed.

You should use = std::move but then disarm the moved-from object so it can be safely deconstructed without freeing any of the pointers which are now property of the moved-to object. Set the moved-from object's interior pointers as it if was an empty container. This will also pleasantly leave the moved-from object valid for reassignment or add calls. Please document this behavior in the comment above the function.

For the 64-bit object: you're not copying the local copyOnWrite flag. And neither is the copy constructor that I wrote. Opps! Maybe they should both be = default to prevent further human error.

lemire:
@WojciechMula  Ping!

What is the status of this PR in  your opinion?

WojciechMula:
@lemire I'll try solve it in the way @ksajme suggested. I didn't have free time last week, my parents visited me.

WojciechMula:
OK, I did the changes you suggested. Hope now it's better.

lemire:
@gssiyankai : want to have a look?

gssiyankai:
LGTM

lemire:
I'll merge and edit AUTHORS

lemire:
I have issued a release.

PR #95, opened 2017-04-25
Fix bitmap copy bug when source is empty

hxfxjun:
Fix bitmap copy bug when source is empty


lemire:
There is no rationale nor a corresponding unit test.

Issue #99, opened 2017-05-03
Improving run_container_equals_array


WojciechMula:
The procedure's code:

```cpp
bool run_container_equals_array(const run_container_t* container1,
                                const array_container_t* container2) {
    if (run_container_cardinality(container1) != container2->cardinality)
        return false;
    int32_t pos = 0;
    for (int i = 0; i < container1->n_runs; ++i) {
        uint32_t run_start = container1->runs[i].value;
        uint32_t le = container1->runs[i].length;

        for (uint32_t j = run_start; j <= run_start + le; ++j) {
            if (pos >= container2->cardinality) {
                return false;
            }
            if (container2->array[pos] != j) {
                return false;
            }
            ++pos;
        }
    }
    return (pos == container2->cardinality);
}
```

In my opinion the inner loop might be replaced by only three comparisons.
Since an array container contains a sorted and unique set of values, it is
sufficient to perform following checks:

```cpp
     
     // ...
     for (int i = 0; i < container1->n_runs; ++i) {
        uint32_t run_start = container1->runs[i].value;
        uint32_t le = container1->runs[i].length;

        /* 1 */ if (pos + le >= container2->cardinality) return false;
        /* 2 */ if (container2->array[pos] != run_start) return false;
                // check 1 guarantees that the array is big enough
        /* 3 */ if (container2->array[pos + le] != run_start + le) return false;
        pos += le;
     }
```

If array values at run's boundaries (pos and pos + le) match the run range,
then all values at index pos + i (i = 1 ... pos + le - 1) must be equal to
run_start + i.

Is there anything I missed?

fsaintjacques:
IIRC this is due to RLE-tuples not being automatically packed, e.g. `[[0, 10], [10, 2]]` is equivalent to `[[0,12]]`.

fsaintjacques:
heh, forget what I said, coffee didn't kick in.

fsaintjacques:
Do we need the first cardinality check? The first condition of the function is that `run.card == array.card`, thus `sum(le) == array.length`.

WojciechMula:
You're right, it's not needed. I also made a mistake, pos has to be update by le + 1.

fsaintjacques:
Anyhow, good find, that should speed up considerably such comparisons.

WojciechMula:
@fsaintjacques Thanks for discussion. I'll try to craft a pull request.

lemire:
@WojciechMula Implemented in Java

[...]

PR #100, opened 2017-05-09
Improved equality of run and array containers

WojciechMula:
This PR implements optimization from #99. I didn't add any new unit test, as existing tests already cover the changed function.


lemire:
Looks good.

PR #102, opened 2017-05-13
Improve equality of run and array containers even further

gssiyankai:


lemire:
This makes a lot of sense.

PR #104, opened 2017-05-16
Fix memory leak in roaring_bitmap_or_many.

Ezibenroc:
When the length of one of the bitmaps was 0, then the bitmap created by `roaring_bitmap_create_with_capacity` was never freed nor returned.


lemire:
Good catch.

PR #105, opened 2017-05-18
Convert to Python3

ralic:
using : find . -name '*.py' | xargs 2to3 -w


Ezibenroc:
Switching to Python3 is a good idea. But maybe you want to remove the backup file created by `2to3`?

PR #108, opened 2017-07-06
Eliminate unneeded calloc() calls

### LINKED ISSUE #65: Optimization opportunity : reduce memset
The current code uses `calloc` for initializing bitset containers, whether it is needed or not. (In several cases, it is not needed, a simple `malloc` would suffice.)

For some operations, this carries a price that shows up in the profiling.

spektom:
As there's no aligned_calloc(), which would allow to eliminate memset() that's more expensive the only way to optimize is use memset() still, but remove unneeded calloc() calls.

This should fix #65


lemire:
Thanks LGTM.

PR #109, opened 2017-07-24
C++/CLI shim (64bit access for .NET callers)

K2:
This is a very minimal shim at the moment I have tested the save/restore add/contains and most of the other basic methods.

If anybody wan'ts more of the exotic functionality available in the JAVA, C# or C++ or other API's let me know and I can formulate the syntax for C++/CLI.

Thanks! @lemire   and others! I'm using this in my project for the moment. 

https://github.com/RoaringBitmap/CRoaring/compare/master...K2:master?expand=1#diff-96f61276013ef22f33eb11ef6785bbdf

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -315,3 +315,4 @@ __pycache__/
 *.pyc
 
 #############END VISUAL STUDIO##############
+/visualstudio
--- a/cpp/export.cpp
+++ b/cpp/export.cpp
@@ -1,50 +1,187 @@
-#define WIN32_LEAN_AND_MEAN
+#pragma unmanaged
 #include "roaring64map.hh"
-#pragma managed
 
-#include <msclr\marshal.h>
+#pragma managed
+#include <vcclr.h> 
+#include <msclr/marshal.h>
 
 using namespace System;
+using namespace System::Collections::Generic;
+using namespace System::Runtime::InteropServices;
 
 namespace RoaringCLI {
-	public ref class Roar
+
+	public ref class RoarCLI
 	{
 	private:
 		Roaring64Map *r64;
-		static Roar() { }
+
+		static RoarCLI() { }
 
 	protected:
-		__clrcall !Roar() { delete r64; }
+		__clrcall !RoarCLI() { delete r64; }
 
-	public: 
-		__clrcall Roar() {
+	public:
+		__clrcall ~RoarCLI() { delete r64; }
+
+		// Managed iterator passed directly to native code
+		delegate bool ManagedIterator(uint64_t value, uint64_t %param);
+		virtual bool __clrcall iterate(ManagedIterator^ iterator, size_t %high_bits) sealed {
+
+			GCHandle gch = GCHandle::Alloc(iterator);
+			IntPtr stubPointer = Marshal::GetFunctionPointerForDelegate(iterator);
+			roaring_iterator64 functionPointer = static_cast<roaring_iterator64>(stubPointer.ToPointer());
+
+			size_t counter = high_bits;
+
+			r64->iterate(functionPointer, &counter);
+			
+			gch.Free();
+			
+			high_bits = counter;
+			return true;
+		}
+
+		// constructors
+		__clrcall RoarCLI() {
 			r64 = new Roaring64Map();
 		}
 
-		__clrcall Roar(Roaring64Map& other) {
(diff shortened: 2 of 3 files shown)


K2:
I forgot to say that there are required settings  to be applied to this file in the roaring project to build properly.  These are the settings for the "export.cpp" file itself, no other settings for the main project are needed to be changed.  The "Basic Runtime Checks" needs to be set to Default (not sure why it's not bold since I did change it) then clearly /clr and compile as C++. That's it!

image

lemire:
Ok. We need someone with a good CLR expertise to review this. Maybe @RogueException ?

K2:
I added a C# unit test that follow's closely to the cpp_unit and also improved error handling and exposed the full set of 64 bit interfaces from the Roaring64Map class.

lemire:
+1

lemire:
Ok. So nobody has jumped in to review this. It is a bit difficult for me because I don't know CLI. Still, I am tempted to merge it on the belief that it cannot cause harm. It is a thin layer of code that cannot be terribly wrong. 

I do have one naive technical question. Your instance operators seem to create new instances. Is this the desired effect?

Ok. On the assumption that all the code is correct and efficient (an assumption that I am willing to make), would you be so kind as to add a tiny bit of commentary. I am thinking about a sentence or two at the start of each of the two files, plus maybe a paragraph or so in the README (maybe in our Visual Studio section). Or, if you prefer, in a distinct file. What I'd like, in an ideal world, is a description allowing someone as naive as I can be with CLI to pick up Visual Studio and run your tests. If that's too much to ask, then, at least, a description that explains to someone more knowledgeable where the files are and how you expect them to use it.

lemire:
Moreover, it would be interesting to relate your work to https://github.com/RogueException/CRoaring.Net

For example, why include the CLI wrapper within the CRoaring library when we already have a C# wrapper called CRoaring.Net? I think this needs to be briefly discussed. 

Any reason why you did not use/extend CRoaring.Net?

K2:
Thanks @lemire I guess I should of merged the thread over here instead of the #1 I think I got to you're points, more or less extending an existing native implementation into .NET, for me, was easier than writing a Tree or Map to segment the underlying 32bit maps (as Roaring64Map does).  If I had realized the slowdown for *Many operators as their currently simulated with loops I think I may of attempted my own implementation earlier :)  CRoaring.Net is great and has coreclr support constraints, I'm not sure if that's impacting his decision to extend or anything else, however I'm open to scraping the dependency on Roaring64Map and ground up implementing Tree or something else that can do better with the *Many call's or forking so it's more clear to consumers how to get at this version etc...

lemire:
> Sure I can take a look at CMake if you'd like 

I guess that this would only make sense in a cross-platform setting. 


> To me it seemed less error prone to add 64 bit .NET support by extending an existing underlying native implementation, I'd be assured not to inject any serious bugs. An added bonus was that as the Roaring64Map class performance improves it'd automatically boost this interface.

I don't think that's in question. The work you did makes sense.

> I can fork this off or invite @RogueException for suggestions or @ksajme or anybody else...

Let us be practical.

I don't think a fork makes sense. You are not changing CRoaring itself, so why fork?

If you look at http://roaringbitmap.org/software/ you will notice that we have several wrappers around CRoaring, for various languages. Could we consider your work as one such wrapper?

You could create a distinct repository (we could put it under https://github.com/RoaringBitmap if you'd like) where you could put your code in a way that is convenient for Visual Studio users.

What about that?

K2:
Sure, that'd be good, yeah list it in the wrappers area, I can make some more doc/cleanup to help out. 

 btw I was reading http://lemire.me/blog/2012/05/31/data-alignment-for-speed-myth-or-reality/ when I was thinking about some of the extensions to the format for 64 bit, do you think it would work out with 24 bit chunks instead of 16 bit?  It wouldn't be aligned so much but playing with some macro's/alloca todo some converting between int/int24/int64 could could be OK.. the sizes would balloon a bit into 16MB size instead of 8kb...

lemire:
@K2 Packing/unpacking 24-bit values would be fast (maybe as fast as packing/unpacking 16-bit values).

From time to time, your unaligned 24-bit value would fit in two cache lines, but that's not likely to be a performance issue.

lemire:
Ok. I am going to close this for now. Ping me back later.

PR #110, opened 2017-08-02
Added AVX support to MSVC builds

jonathanmarston:
This change adds support for builds that use AVX extensions using Microsoft Visual C++ 2015 or higher.

### DIFF
--- a/include/roaring/portability.h
+++ b/include/roaring/portability.h
@@ -38,14 +38,17 @@
 
 #if defined(_MSC_VER)
 #define __restrict__ __restrict
-#if _MSC_VER > 1800
-#define __AVX2__
-#define __BMI2__
-#endif
 #endif
 
 #ifndef DISABLE_X64  // some users may want to compile as if they did not have
                      // an x64 processor
+
+#if defined(_MSC_VER)
+#if _MSC_VER > 1800 & defined(_M_X64)
+#define __AVX2__
+#define __BMI2__
+#endif
+#endif
 
 // unless DISABLEAVX was defined, if we have AVX2 and BMI2, we enable AVX
 #if (!defined(USEAVX)) && (!defined(DISABLEAVX)) && (defined(__AVX2__)) && \
--- a/include/roaring/portability.h
+++ b/include/roaring/portability.h
@@ -38,6 +38,10 @@
 
 #if defined(_MSC_VER)
 #define __restrict__ __restrict
+#if _MSC_VER > 1800
+#define __AVX2__
+#define __BMI2__
+#endif
 #endif
 
 #ifndef DISABLE_X64  // some users may want to compile as if they did not have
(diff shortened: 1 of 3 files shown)


lemire:
Thanks. Let us review this soon.

lemire:
Thanks for bringing this issue up. 

Visual Studio does define __AVX2__ when setting  /arch:AVX2. This is documented. Forcing __AVX2__ whenever an x64 architecture is detected would produce faulty binaries on hardware that does not support __AVX2__. Moreover, your proposed change would make life difficult for Visual Studio users who wish to build an x64 version of the code without AVX2 support.

A better option is for the programmer to add the /arch:AVX2 compiler option when he or she determines that it is safe to do so. I have also made sure that if you call CMake for Visual Studio with "-DFORCE_AVX", then __AVX2__ will be defined (this requires no change to the C source code).

The BMI support is only useful if you also use inline assembly but Visual Studio does not support inline assembly make the determination regarding BMI support moot.

I have updated the documentation...

https://github.com/RoaringBitmap/CRoaring#building-visual-studio-under-windows

I am going to close this issue. Please reopen if you feel that my changes do not meet your needs.

Thanks again.

jonathanmarston:
I agree that manually defining __AVX2__ in a header is not ideal, but I was not having (and still not having) any luck with getting CMake to pass /arch:AVX2 to the compiler. Even with your suggested parameter in CMakeLists.txt, __AVX2__ isn't showing as defined in VS 2017 on my machine.

The bigger issue is that the code in portability.h will only define USEAVX if __BMI2__ is also defined (and in fact explicitly undefs it if __BMI2__ isn't defined). Since the MSVC compiler does in fact support BMI2 intrinsics, I felt it was better to go ahead and define __BMI2__ and update the code in utilasm.h to not assume that its ok to use inline asm just because BMI is supported.

I've confirmed that simply adding the parameter that you suggest doesn't actually enable AVX optimizations by benchmarking a build with the code in master and -DFORCE_AVX2=ON vs a build with the code from my pull request. With the changes in this pull request bitset operations run over 6x faster.

lemire:
You are correct that we would require both AVX2 and BMI2 for some reason, but I cleaned this up.

Note that the flag is  -DFORCE_AVX=ON and not  -DFORCE_AVX2=ON.

See this screenshot... 

https://goo.gl/photos/XH7peTKYRCSxWzph9

Reopen if this does not solve your issue.

lemire:
I have substantially improved the relevant part of the README file: https://github.com/RoaringBitmap/CRoaring

I am hopeful that this should address your concerns.

jonathanmarston:
With the combination of the last set of changes that you made to portability.h and using the correct flag in my CMake settings (oops!), I'm now getting an AVX-enabled build out of the master branch. Thanks!

lemire:
Fantastic.

Note that Microsoft VSTS uses Roaring bitmaps...

https://blogs.msdn.microsoft.com/congyiw/2015/12/14/why-does-cloning-from-vsts-return-old-unreferenced-objects/

PR #124, opened 2018-02-15
Sync inline documentation for iterate C++ methods

WojciechMula:


lemire:
Great. Thanks for the contribution.

PR #132, opened 2018-04-05
Disables including malloc.h on FreeBSD, where it doesn't exist

yuce:
On FreeBSD 11 the compilation fails with:
```
/usr/include/malloc.h:3:2: error: "<malloc.h> has been
      replaced by <stdlib.h>"
```
This PR disables including `malloc.h` on FreeBSD, which makes `make` and `make test` succeed.


lemire:
Thanks. I’d do away with malloc.h but some people report that it is needed.

Anyhow, great!

yuce:
Thanks!

PR #140, opened 2018-05-24
Fixed infinite loop in roaring_bitmap_from_range()

andreigudkov:
Range [0, 2^32) caused integer overflow and infinite loop.


lemire:
Great.

PR #143, opened 2018-05-28
cmake: add ROARING_ namespace to all cmake variables

senior7515:
* add cmake namespacing
* update docs


senior7515:
without cmake namespacing, variables such as `BUILD_STATIC` might control multiple projects at a time. This is necessary to integrate into larger cmake projects.

senior7515:
@lemire  i just pushed the latest fixes, running the all the tests suites including pre-release.sh

amalgamation.sh works too. Is there anything else I need to check?

senior7515:
btw, no need to search through the code as the macro definitions are gated **by** the cmake variables. 

`ROARING_DISABLE_AVX` gates the code definitions `-DDISALBEAVX` as far as I could tell.

senior7515:
in particular these variables 
```
if(NOT MSVC)
if(NOT ROARING_DISABLE_NATIVE)
set(OPT_FLAGS "-march=native")
endif()
endif()

if(ROARING_DISABLE_X64)
  # we can manually disable any optimization for x64
  set (OPT_FLAGS "${OPT_FLAGS} -DDISABLE_X64" )
endif()
if(ROARING_DISABLE_AVX)
  # we can manually disable AVX by defining DISABLEAVX
  set (OPT_FLAGS "${OPT_FLAGS} -DDISABLEAVX" )
endif()

if(FORCE_AVX) # some compilers like clang do not automagically define __AVX2__ and __BMI2__ even when the hardware supports it
if(NOT MSVC)
   set (OPT_FLAGS "${OPT_FLAGS} -mavx2 -mbmi2")
else()
   set (OPT_FLAGS "${OPT_FLAGS} /arch:AVX2")
endif()
endif()


```

senior7515:
all the test combinations seemed to have worked i.e.: 

```
tools/prereleasetests.sh 
```

lemire:
Good.

lemire:
Merging.

PR #145, opened 2018-06-01
Add all functions for function containsRange()

GuillaumeHolley:


lemire:
Fantastic. Will review.

lemire:
Ok. This looks good and I am going to merge this. However, before we release this, we need unit tests. We can't release an untested function.

@GuillaumeHolley Would you consider throwing a few tests?

You could add a function or two there:

https://github.com/RoaringBitmap/CRoaring/blob/master/tests/toplevel_unit.c

GuillaumeHolley:
@lemire Thanks for having a look at the code. I will make some unit tests in there today.

@richardstartin Well, I didn't check the whole library but based on this function and its comment:

`/**
 * compute (in place) the negation of the roaring bitmap within a specified
 * interval.
 * areas outside the range are passed through unchanged.
 */

void roaring_bitmap_flip_inplace(roaring_bitmap_t *x1, uint64_t range_start,
                                 uint64_t range_end);`

I assumed I could go for all inclusive. 

The real reason I went for all inclusive is that functions of type `*_container_contains()` take as parameters arguments of type `uint16_t` which means that if I wanted to preserve that type to match the other contain functions, I couldn't use exclusive upper bounds.

andreigudkov:
I am implementing `add_range()` function and also need to decide whether to make new functions with inclusive or exclusive ranges. Let's agree on some uniform style. Existing code uses exclusive ranges, so I think it is better to stick to it. The downside is that exclusive ranges require to use wider types.

richardstartin:
@GuillaumeHolley I'm not at all familiar with this library, and don't expect that it should be consistent with the Java version, so I'm not saying inclusive intervals are wrong, just surprising to me.

GuillaumeHolley:
In the exisiting code, `roaring_bitmap_flip_inplace()` seems to use inclusive bounds but I just noticed that `roaring_bitmap_from_range()` uses exclusive upper bound. In my opinion, I think we should stick/go for inclusive bounds. It might be surprising but as said @andreigudkov, exclusive upper bounds forces us to use wider types for all range functions while inclusive bounds do not have this issue. As using inclusive or exclusive bounds does not change anything for the user, I think we should go for the solution which does not have a penalty (using wider types). Of course, I will adapt to what is decided.

@richardstartin I perfectly understand your point and I never assumed you were expecting the library to be 100% consistent with the Java version, I apologize if I gave you that impression. I was trying to explain the logic that motivated my choice for inclusive bounds.

lemire:
> In the existing code, roaring_bitmap_flip_inplace() seems to use inclusive bounds but I just noticed that roaring_bitmap_from_range() uses exclusive upper bound. 

They both use exclusive upper range.

It "feels" easier to use inclusive ranges, and it is certainly sensible, but there are downsides. 

For example, if you do "add_range(x,0,400)", then how many values have you added? You have added 400 or 399? If you use exclusive upper ranges, then it is 400, otherwise 399.

The point is that if you write your intervals as [a,b), the length of the interval is b-a. If you write them as [a,b], then the length is b-a+1.  

The typical way we write a for loop in C is...

```
for (int i = start; i < end ; i++)
```

In python, you have

```
for i in range(1,3):
  print(i)
```

will print 1,2 and not 3.

The annoying thing is that to cover the full 32-bit range, with an exclusive upper bound, you need to use 64-bit ints.

In C++ STL, the "end" is always unreachable, so exclusive:

```
 fill(v.begin(), v.begin() + 10, 3 );
```

This writes 10 values, not, as you might expect, 11...


To be clear, I would prefer inclusive range... it makes more sense to me logically... but the trouble is that socially, many people expect exclusive upper ranges.

Note that we can support both. Just throw in convenience functions. You just have to make sure that they have different names so that confusion is not possible.

lemire:
Last commit clarifies the semantics of the flip functions:

https://github.com/RoaringBitmap/CRoaring/commit/7c757a54a3b46e3e3e70dcdf03ca7382fc48f057

lemire:
Note: please consider that whether the final value is excluded or included is really, ultimately, [bikeshedding](https://en.wiktionary.org/wiki/bikeshedding). If one implements one of them, the other one follows trivially.

richardstartin:
> Note: please consider that whether the final value is excluded or included is really, ultimately, bikeshedding. 

Yes, I agree, in fact I'm sorry I brought it up!

lemire:
@richardstartin 

No!!! I did not want to be dismissive of your comment. I just mean that it should not hold out software work. It is important in the way details are important. It can trip people off... but it is not hard engineering.

lemire:
Ping @GuillaumeHolley : Can you let us know whether you plan to throw in some tests? If the answer is no, I will merge without tests and strap in the tests afterward.

It is just nicer to merge a tested PR.

GuillaumeHolley:
@lemire : Sorry for the delay, I am making the necessary modifications for the excluded upper bound at the moment and will throw in some tests right afterwards. Should be ready today.

lemire:
See https://github.com/RoaringBitmap/CRoaring/blob/master/AUTHORS

PR #147, opened 2018-06-11
Clarify Windows C++ support.

cjmay:
This commit adds a caveat to the README that C++ is not supported on Microsoft Visual Studio.  Between the example using CRoaring from C++ later in the README and these statements toward the beginning, I expected Visual C++ to work (observing it supports C++11), and thought it was a bug that it did not:

> Portable Roaring bitmaps in C (and C++) with full support for your favorite compiler (GNU GCC, LLVM's clang, Visual Studio). 

> . . . there is also an optional C++ class that requires a C++ compiler supporting the C++11 standard.

However, I tried to be conservative/unobtrusive; the Visual Studio support is hopefully now explicit without detracting from the README as a whole.

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,9 +48,9 @@ of the latest hardware. Roaring bitmaps are already available on a variety of pl
 # Requirements
 
 - The library should build on a  Linux-like operating system (including MacOS).
-- We also support Microsoft Visual studio.
+- We also support Microsoft Visual Studio (C only).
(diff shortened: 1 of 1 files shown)


lemire:
As stated, this is wrong. Not only is C++ supported under Visual Studio, it is actually tested as part of the CI:

https://ci.appveyor.com/project/lemire/croaring/build/job/vmuwgtg98f8fs78f

If you have encountered some issue, please open a specific issue.

Visual Studio failing to compile our C++ code would be a bug.

PR #148, opened 2018-06-13
New function: roaring_bitmap_add_range()

andreigudkov:


lemire:
Please update bitset_range_cardinality to bitset_lenrange_cardinality. As you pointed out, bitset_range_cardinality  is buggy and it was dead code.

lemire:
Or else, if you want bitset_range_cardinality, then your PR should provide its own (non-buggy) version. 

I recommend that you pull from origin https://github.com/RoaringBitmap/CRoaring since I made appveyor more verbose which could help diagnose problem. Your PR mysteriously failed on Win32...

lemire:
LGTM. Merging.

PR #153, opened 2018-06-20
Implement ordered_writer and roaring_bitmap_add_ordered for faster construction

telles-simbiose:
This tries to port OrderedWriter from Java implementation (the behaviour of ordered_writer.c is quite the same).
This implements a new function on roaring.h: `roaring_bitmap_add_ordered` which allocates and uses an ordered_writer.

There is still work to do which I need help (this is my first C "serious" code, so there might be other mistakes).

### DIFF
--- a/include/roaring/ordered_writer.h
+++ b/include/roaring/ordered_writer.h
@@ -17,6 +17,6 @@ void roaring_bitmap_writer_free(roaring_bitmap_writer_t *writer);
 
 void roaring_bitmap_writer_flush(roaring_bitmap_writer_t *writer);
 
-void roaring_bitmap_writer_add(roaring_bitmap_writer_t *writer, const uint32_t val);
+bool roaring_bitmap_writer_add(roaring_bitmap_writer_t *writer, const uint32_t val);
 
 #endif
--- a/include/roaring/ordered_writer.h
+++ b/include/roaring/ordered_writer.h
@@ -1,0 +1,18 @@
+#include <assert.h>
+#include <roaring/roaring.h>
+
+typedef struct roaring_bitmap_writer_s {
+    uint64_t *bitmap;
+    roaring_bitmap_t *target;
+    uint32_t current_key;
+    uint64_t cardinality;
+    bool dirty;
+} roaring_bitmap_writer_t;
+
+roaring_bitmap_writer_t *roaring_bitmap_writer_create(roaring_bitmap_t *target);
+
+void roaring_bitmap_writer_free(roaring_bitmap_writer_t *writer);
+
+void roaring_bitmap_writer_flush(roaring_bitmap_writer_t *writer);
+
+void roaring_bitmap_writer_add(roaring_bitmap_writer_t *writer, const uint32_t val);
--- a/include/roaring/ordered_writer.h
+++ b/include/roaring/ordered_writer.h
@@ -13,6 +13,8 @@ typedef struct roaring_bitmap_writer_s {
 
 roaring_bitmap_writer_t *roaring_bitmap_writer_create(roaring_bitmap_t *target);
 
+void roaring_bitmap_writer_set(roaring_bitmap_writer_t *result, roaring_bitmap_t *target);
+
 void roaring_bitmap_writer_free(roaring_bitmap_writer_t *writer);
 
 void roaring_bitmap_writer_flush(roaring_bitmap_writer_t *writer);
--- a/include/roaring/ordered_writer.h
+++ b/include/roaring/ordered_writer.h
@@ -5,7 +5,6 @@ typedef struct roaring_bitmap_writer_s {
     uint64_t *bitmap;
     roaring_bitmap_t *target;
     uint32_t current_key;
-    uint64_t cardinality;
     bool dirty;
 } roaring_bitmap_writer_t;
 
--- a/include/roaring/ordered_writer.h
+++ b/include/roaring/ordered_writer.h
@@ -1,3 +1,6 @@
+#ifndef ROARING_ORDERED_WRITER
+#define ROARING_ORDERED_WRITER
+
 #include <assert.h>
 #include <roaring/roaring.h>
 
@@ -15,3 +18,5 @@ void roaring_bitmap_writer_free(roaring_bitmap_writer_t *writer);
 void roaring_bitmap_writer_flush(roaring_bitmap_writer_t *writer);
 
 void roaring_bitmap_writer_add(roaring_bitmap_writer_t *writer, const uint32_t val);
+
+#endif
--- a/include/roaring/roaring.h
+++ b/include/roaring/roaring.h
@@ -11,6 +11,7 @@ extern "C" {
 #include <roaring/roaring_array.h>
 #include <roaring/roaring_types.h>
 #include <roaring/roaring_version.h>
+#include <roaring/ordered_writer.h>
 #include <stdbool.h>
 
 typedef struct roaring_bitmap_s {
@@ -225,6 +226,13 @@ void roaring_bitmap_free(roaring_bitmap_t *r);
  */
 void roaring_bitmap_add_many(roaring_bitmap_t *r, size_t n_args,
                              const uint32_t *vals);
+
+/**
+ * Add ordered values from pointer vals, faster than repeatedly calling
+ * roaring_bitmap_add_many for ordered values
+ *
+ */
+void roaring_bitmap_add_ordered(roaring_bitmap_t *r, size_t n_args, const uint32_t *vals, roaring_bitmap_writer_t *writer);
 
 /**
  * Add value x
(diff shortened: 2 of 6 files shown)


telles-simbiose:
@lemire , there is a memory leak somewhere in this code which I couldn't find. Can you take a look?
Thank you so much

lemire:
Thanks. At a glance, this looks reasonable to me. No obvious problem.  

I'm currently traveling so it is difficult for me to review code.

 @andreigudkov would you volunteer to have a look?

src/roaring.c
```diff
@@ -81,6 +82,35 @@ roaring_bitmap_t *roaring_bitmap_create_with_capacity(uint32_t cap) {
     return ans;
 }
 
+
+void roaring_bitmap_add_ordered(roaring_bitmap_t *r, size_t n_args, const uint32_t *vals) {
```
Ezibenroc: Maybe you should make `roaring_bitmap_add_ordered` return an error code, e.g., `1` if the values are ordered and `0` if they are not.

Ezibenroc:
There are a few issues in your code that you should fix.

You need to add `ordered_writer.c` in [src/CMakeLists.txt](https://github.com/RoaringBitmap/CRoaring/blob/master/src/CMakeLists.txt#L34), in the definition of `ROARING_SRC`.

You also need to add `$SCRIPTPATH/include/roaring/ordered_writer.h` in [amalgamation.sh](https://github.com/RoaringBitmap/CRoaring/blob/master/amalgamation.sh#L40), at the end of the definition of `ALLCHEADERS`.

Finally, it would be great to add some tests.

lemire:
@Ezibenroc +1

telles-simbiose:
Thanks so much @lemire @Ezibenroc @richardstartin for your contributions.
These compilations issues must be because I was writing this code in another project.

I am most concerned about the memory leak, I ran out of memory after running this in a loop: create a roaring bitmap, add values with this function, free the roaring bitmap

include/roaring/roaring.h
```diff
@@ -226,6 +227,13 @@ void roaring_bitmap_free(roaring_bitmap_t *r);
 void roaring_bitmap_add_many(roaring_bitmap_t *r, size_t n_args,
                              const uint32_t *vals);
 
+/**
+ * Add ordered values from pointer vals, faster than repeatedly calling
+ * roaring_bitmap_add_many for ordered values
+ *
+ */
+void roaring_bitmap_add_ordered(roaring_bitmap_t *r, size_t n_args, const uint32_t *vals, roaring_bitmap_writer_t *writer);
```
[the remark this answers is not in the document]
telles-simbiose: It was actually `void roaring_bitmap_add_ordered(roaring_bitmap_t *r, size_t n_args, const uint32_t *vals);`
Fixed

include/roaring/roaring.h
```diff
@@ -11,6 +11,7 @@ extern "C" {
 #include <roaring/roaring_array.h>
 #include <roaring/roaring_types.h>
 #include <roaring/roaring_version.h>
+#include <roaring/ordered_writer.h>
```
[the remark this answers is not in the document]
telles-simbiose: Fixed

src/ordered_writer.c
```diff
@@ -0,0 +1,62 @@
+#include <assert.h>
+#include <roaring/roaring.h>
```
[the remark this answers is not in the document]
telles-simbiose: Fixed

src/roaring.c
telles-simbiose: roaring_bitmap_writer_add() has an assert() for this

Ezibenroc: I saw, but this is a problem since we cannot "catch" it. In case of invalid input, the program will crash, so the user has to check that the operation is valid beforehand.

This is similar to [`roaring_bitmap_select`](https://github.com/RoaringBitmap/CRoaring/blob/master/include/roaring/roaring.h#L515-L521): this function returns true when the operation is valid, false otherwise.

src/roaring.c
```diff
@@ -81,6 +82,35 @@ roaring_bitmap_t *roaring_bitmap_create_with_capacity(uint32_t cap) {
...
+
+    if (n_args == 0) {
+        return;
+    }
+    if (n_args == 1) {
+        roaring_bitmap_add(r, vals[0]);
+        return;
+    }
+
+    if (n_args < 1024 || r->high_low_container.size > 0) {
```
andreigudkov: I do not entirely understand this condition. Why do we need `r->high_low_container_size > 0`? And why is 1024 used as threshold and not some other value?

include/roaring/ordered_writer.h
```diff
@@ -0,0 +1,18 @@
...
+    bool dirty;
+} roaring_bitmap_writer_t;
+
+roaring_bitmap_writer_t *roaring_bitmap_writer_create(roaring_bitmap_t *target);
+
+void roaring_bitmap_writer_free(roaring_bitmap_writer_t *writer);
+
+void roaring_bitmap_writer_flush(roaring_bitmap_writer_t *writer);
+
+void roaring_bitmap_writer_add(roaring_bitmap_writer_t *writer, const uint32_t val);
```
andreigudkov: This function should be inlined.

include/roaring/ordered_writer.h
```diff
@@ -0,0 +1,18 @@
+#include <assert.h>
+#include <roaring/roaring.h>
+
+typedef struct roaring_bitmap_writer_s {
+    uint64_t *bitmap;
```
andreigudkov: This implementation is the exact copy of bitmap_container_t. I think it is better to replace `bitmap` and `cardinality` fields with bitmap_container_t* instance. This will also allow us to make `roaring_bitmap_writer_flush()` more efficient (see other comments).

src/roaring.c
telles-simbiose: Got it! I made it return true indicating success.

telles-simbiose:
That's odd. My builds on Travis are failing on tests that my code shouldn't interfere (I didn't make any test to call this new feature yet)

src/ordered_writer.c
```diff
@@ -0,0 +1,62 @@
...
+
+    bitset_container_t *newac = bitset_container_create();
+
+    newac->cardinality = -1;
+    memcpy(newac->array, bitmap, sizeof(uint64_t) * BITSET_CONTAINER_SIZE_IN_WORDS);
+
+    uint8_t new_typecode = BITSET_CONTAINER_TYPE_CODE;
+    void *newcontainer = container_repair_after_lazy(newac, &new_typecode);
+
+    ra_append(&target->high_low_container, writer->current_key, newcontainer, new_typecode);
```
andreigudkov: This will work only if roaring bitmap was empty.

Instead we can do the following:
1. Redefine `roaring_bitmap_writer_s` as following:

       typedef struct roaring_bitmap_writer_s {
         roaring_bitmap_t *target;
         bitmap_container_t* tmp_container;
         uint16_t current_key;
       };
2. Every time `roaring_bitmap_writer_add()` is called, add given value directly into `tmp_container`
3. When time comes to flush, get container from `writer->target` with the key `writer->current_key`
    * If such container exists, then call `container_ior` with `tmp_container`
    * If it doesn't exist, then insert `tmp_container` into roaring_bitmap.

src/roaring.c
telles-simbiose: Sorry about this `n_args < 1024`, it shouldn't be there.
`r->high_low_container.size > 0` is because this implementation only works when the roaring_bitmap is empty

include/roaring/ordered_writer.h
telles-simbiose: I actually removed `cardinality`, it was there just because I was using the same code as bitset_container_t

src/ordered_writer.c
telles-simbiose: The point is to avoid inserting right in the containers for every value to avoid reallocation, but it's possible to achieve this If I create bitset_container_t and use this logic, right?

andreigudkov:
How do you test for memory leaking? Do you have unit tests or sample app that demonstrates that?

telles-simbiose:
The memory leak was on my application, sorry for bugging you guys!

lemire:
@telles-simbiose 

You are supposed to have the function declaration and definition in a header file (*.h):

      inline bool roaring_bitmap_writer_add(roaring_bitmap_writer_t *writer, 
       const uint32_t val) {...}

Then put the following in some soure (*c) file:

      extern inline bool roaring_bitmap_writer_add(roaring_bitmap_writer_t *writer, 
       const uint32_t val);

lemire:
@telles-simbiose

Please note my last comment and take notice of the appveyor failure. Your PR is still incorrect. You need to apply a minor fix.

lemire:
Ping! @telles-simbiose

feng-y:
ping @telles-simbiose

lemire:
This has been opened for too long without progress. Closing.

Issue #158, opened 2018-06-22
add checked, remove checked


SalvatorePreviti:
At the moment "add" and "remove" operations don't return an indication wether the operation changed the bitmap or not.

It is however useful in several scenarios to have something like this:

```
if (!roaring_bitmap_contains(roaring, value)) {
   roaring_bitmap_add(roaring, value);
   do_something_very_important_only_once(value);
}
```

The Java version has two methods "checkedAdd" and "checkedRemove".

Would make sense to provide two new functions "roaring_bitmap_add_checked" and "roaring_bitmap_remove_checked" that perform better than doing first the "has" operation or checking the cardinality of the whole bitmap after the operation?

Example:
```
if (roaring_bitmap_add_checked(bitmap, value)) {
   do_something_very_important_only_once(value);
}
```

SalvatorePreviti:
Here a pull request - https://github.com/RoaringBitmap/CRoaring/pull/159

SalvatorePreviti:
Merged.

PR #159, opened 2018-06-22
add_checked_and_remove_checked

SalvatorePreviti:
https://github.com/RoaringBitmap/CRoaring/issues/158

At the moment "add" and "remove" operations don't return an indication wether the operation changed the bitmap or not.

It is however useful in several scenarios to have something like this:

```
if (!roaring_bitmap_contains(roaring, value)) {
   roaring_bitmap_add(roaring, value);
   do_something_very_important_only_once(value);
}
```

The Java version has two methods "checkedAdd" and "checkedRemove".

Would make sense to provide two new functions "roaring_bitmap_add_checked" and "roaring_bitmap_remove_checked" that perform better than doing first the "has" operation or checking the cardinality of the whole bitmap after the operation?

Example:
```
if (roaring_bitmap_add_checked(bitmap, value)) {
   do_something_very_important_only_once(value);
}
```


lemire:
LGTM.

PR #161, opened 2018-06-27
fix ra_clear, it should reset size and allocated size

SalvatorePreviti:
After calling ra_clear the state of roaring_array_t may be not correct and this can break the usage of other functions on the cleared array.

ra_clear already resets all pointers to NULL, it is safer to reset also `size` and `allocation_size` to zero.

It is then possible to call other functions on the cleared array without any risk.

### DIFF
--- a/src/roaring_array.c
+++ b/src/roaring_array.c
@@ -191,10 +191,12 @@ void ra_reset(roaring_array_t *ra) {
 }
 
 void ra_clear_without_containers(roaring_array_t *ra) {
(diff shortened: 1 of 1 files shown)


lemire:
Sure.

Note that the "bug" label is maybe too much.

My threshold for a bug is some behaviour that would negatively affect the users of the public API library. I don't think we have this now.  Anyone down at the ra_clear level better know what he/she is doing.
Cheers!

PR #170, opened 2018-07-08
Avoid warnings when sizeof(size_t) < sizeof(uint64_t)

madscientist:
Sorry, I should have ensured the 32bit build was completely successful before the last pull.  After this, it is.

I'm not sure whether it's worthwhile to check that `sizeasarray` would fit into a `size_t` and fail somehow if not; I couldn't see any defined way to return an error.  It's highly unlikely that such a thing could be possible, I suspect.


lemire:
Great.

PR #171, opened 2018-07-14
Windows: Use #define for __builtin_unreachable()

madscientist:
If we use inline functions then compile with /Ob0 (disable inline), as
many debug builds will do, this method will no longer have its
intended effect since the __assume(0) will be hidden inside the (not
inlined) function.


lemire:
Rerunning the CI.

lemire:
We are good now.

PR #172, opened 2018-07-18
Added remove_range() function

andreigudkov:


lemire:
Let us try to merge. There is some flakiness in the CI, but I am hopeful I fixed it.

PR #173, opened 2018-07-31
Improvements to allocation policy of array structs

andreigudkov:
array_container_add(): check that array must be converted to bitset
_before_ adding value, not after. This allows to limit array capacity
with DEFAULT_MAX_SIZE most of the time.

array.c: do not allocate space for more than DEFAULT_MAX_SIZE values
(soft limit, which can be exceeded in negation and some pairwise
operations; in this case we try to limit capacity with 64K)

roaring_array.c: do not allocate space for more than MAX_CONTAINERS

### DIFF
--- a/include/roaring/containers/array.h
+++ b/include/roaring/containers/array.h
@@ -161,15 +161,13 @@ static inline int32_t array_container_serialized_size_in_bytes(int32_t card) {
 }
 
 /**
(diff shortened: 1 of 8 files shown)


lemire:
+1

lemire:
Looks good.

lemire:
Merged.

Issue #176, opened 2018-08-03
Allow `march=` target to be specified during build


kanak:
Our build servers have slightly different CPUs from our production machines, and our production machines themselves are varied. We've started disabling `march=native` after we had a `SIGILL` crash when running a binary built on a machine with a newer CPU. 

It would be better if we could instead specify the value of `march`. In our case, we would pass `march=westmere` which is the oldest CPU that we use.

I'm happy to submit a patch to implement this feature, but I'm not sure if you would like this to replace the existing functionality, or if you would like the two to work together.

Thanks!

lemire:
Sounds good. PR invited.

lemire:
Fixed.

Issue #113, opened 2017-09-28
Implement a dispatched model for compilation of the CPU capabilities


K2:
Since we have a range of optimizations currently available and likely to add more/improve over time, the complexity of building with the appropriate configuration can be tricky.  

Ideally, we would have a dispatching library that is able to select our optimal implementation at run time based on the CPU feature set.  Most libraries I’ve seen that offer dispatched models use environment variables to disable or adjust things on the fly.

For GCC we could use https://gcc.gnu.org/wiki/FunctionMultiVersioning, MSVC is a bit more complicated and it may be inevitable that there will be some function pointers to glue things together, I’ve started with some intrinsic routines to get going also and will be evaluating some options in keeping with pure C (no templates).  It'd be nice to have a simple pattern that worked for all compiler/build targets.

lemire:
I think that it is desirable and quite interesting. How to make it work portably is, of course, the main difficulty.

lemire:
Google Pik has SIMD-based runtime dispatch: 

https://github.com/google/pik/blob/master/simd/dispatch.cc

K2:
I took a look and it seems they have a pretty similar impression that most who go this route use either some intrinsic glue or C++ templates.  About 80% of the dispatched code I see is setup with templates.  Is some very light C++ use OK for us or should we always consider pure C?  I was even looking at the possibility of using some JIT from clang to glue up the appropriate callout's ;)...

The majority of pik's design doc seems to be here, https://github.com/google/pik/tree/master/simd, down in the prior designs section they talk about some of their reasoning and reduced CPU support envelope.

I adapted the Linux test cases to Windows and was evaluating them under some of the Intel profiles (seems they have a give away for open source projects if you look also..)

This run may of had profiling hooks slowing it down a tad, but I think I recall you mentioning another set of test cases that may have more coverage?  I was not sure if these were them or not, I had to expand some of them to generate more data as some of tap's didnt register w/o some big-data-esq levels of execution I guess ;)

C:\Users\files\Source\git\RoaringBit\build\benchmarks\Release>real_bitmaps_benchmark.exe C:\Users\files\Source\git\RoaringBit\benchmarks\realdata\
Loaded 1800 bitmaps from directory C:\Users\files\Source\git\RoaringBit\benchmarks\realdata\
Creating 1800 bitmaps took 748652575 cycles
Copying and freeing 1800 bitmaps took 6867095 cycles
[...]

PR #177, opened 2018-08-06
Add ROARING_ARCH build option to control march=

kanak:
As outlined in https://github.com/RoaringBitmap/CRoaring/issues/176 ,
passing `march=native` can cause SIGILL when the build server is newer
than the server running the binary.

Currently, the only way around this issue is to use
`ROARING_DISABLE_NATIVE` and not use `march` at all.

This commit gives users additional control by adding a new CMake build
option to specify the architecture, e.g. `-DROARING_ARCH=westmere`.
If the existing `ROARING_DISABLE_NATIVE` flag is set to `OFF`, then
this new option has no effect.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -12,6 +12,8 @@ set(ROARING_LIB_NAME roaring)
 
 option(ROARING_DISABLE_X64 "Forcefully disable x64 optimizations even if hardware supports it (this disables AVX) " OFF)
 option(ROARING_DISABLE_AVX "Forcefully disable AVX even if hardware supports it" OFF)
+option(ROARING_DISABLE_NATIVE "Forcefully disable -march optimizations" OFF)
+set(ROARING_ARCH "native" CACHE STRING "If ROARING_DISABLE_NATIVE is OFF, the architecture to optimize for (-march)")
 
 IF(${CMAKE_SYSTEM_PROCESSOR} MATCHES "arm")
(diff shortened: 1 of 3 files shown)


lemire:
Looks good, I'll let the CI finish before merger.

andreigudkov:
Why do we need more options? I do not understand. Optimization options are compiler dependent and IMO should not be set in cmake. I would suggest to _remove_ ROARING_DISABLE_AVX and ROARING_DISABLE_NATIVE instead of adding third (!) option.

See also: https://github.com/RoaringBitmap/CRoaring/issues/113#issuecomment-399911640

andreigudkov:
What I suggest is the following:

 1. Remove entirely ROARING_DISABLE_AVX and ROARING_DISABLE_NATIVE options
 2. For conditional compilation, use `#ifdef __AVX2__`
 3. To select target architecture, use CMAKE_C_FLAGS

kanak:
I wonder if the travis build failure is unrelated? `make test` is consistently passing for me locally.

lemire:
@andreigudkov Unless you object, I will merge this PR. I'd suggest opening a distinct issue regarding the point you rise. Agreed?

andreigudkov:
OK

lemire:
Merging. Not that this merge should not be taken as a vote against @andreigudkov 's objections.

PR #179, opened 2018-08-29
add pagination when getting int array

baiyunping333:


lemire:
Looks good. Can you...

1. Add unit tests...
2. Document ``roaring_bitmap_range_uint32_array`` in ``roaring.h`` so users know what it does?

baiyunping333:
how about! if merge to the `RoaringBitmap/CRoaring`, so I can push to my "redis-roaring"  to [aviggiano/redis-roaring](https://github.com/aviggiano/redis-roaring)

lemire:
I merged just now.

.gitignore
```diff
@@ -317,4 +317,10 @@ paket-files/
 __pycache__/
 *.pyc
 
+amalgamation_demo.c
+amalgamation_demo.cpp
+roaring.c
```
amosbird, line 322: Why do these three files ignored?

lemire: They are automatically generated files that are not part of our source files. See the amalgamation script.

amosbird: Oh, the amalgamated file is $ROOT/roaring.c instead of src/roaring.c. The .gitignore file functions correctly in Git, but tools like fd and ripgrep somehow also ignore src/roaring.c. It will work if we explicitly specify them like this in .gitignore:

```
/amalgamation_demo.c
/amalgamation_demo.cpp
/roaring.c
/roaring.h
/roaring.hh
```

Dr-Emann: That's a good point, it actually only works in gitignore because git only considers the gitignore when it would _add_ a file to be tracked, changes to existing files already in git are tracked regardless of what's in .gitignore.

PR #187, opened 2018-11-02
Software implementation of popcnt

rodenweller:
POPCNT support was not available on processors until around the time of SSE4.2 instructions. CRoaring flag for SSE4 guarantees POPCNT support. This patch uses that flag to guard against attempted usage
of the popcnt instruction on builds where we have not enabled that level of processor support. A software implentation is provided that should provide good all around performance.

### DIFF
--- a/include/roaring/portability.h
+++ b/include/roaring/portability.h
@@ -140,6 +140,7 @@ static inline int __builtin_clzll(unsigned long long input_num) {
 }
 
 /* result might be undefined when input_num is zero */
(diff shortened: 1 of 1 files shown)


lemire:
Restarting the build.

rodenweller:
@lemire Thanks, it looks like the checks passed.

lemire:
Indeed.

lemire:
Published as 0.2.56.

PR #190, opened 2018-11-08
Implement backward iterator

Brian-Esch:
Noteable changes:
- `in_run_index` removed from the `roaring_uint32_iterator_t` which
  breaks abi compatability. Just use the iterator's current value to
  and `run_index` figure out if it should advance or not

- added invariant to `roaring_uint32_iterator_t`: the `container_index`
  can be -1 in addition to `>= parent->high_low_container.size`.
  Previously when the `container_index >= parent->high_low_container_size`
  this represented that the iterator pointed past the end of the
  container. This is still true, however -1 now indicates that the
  iterator points before any of the values in container

- add `roaring_previous_uint32_iterator` method, which moves the
  iterator backwards

- add `roaring_init_iterator_last` method, which initializes the
  iterator to point to the last element in the bitmap

- add prefix and postfix `operator--` to c++ api

### DIFF
--- a/cpp/roaring.hh
+++ b/cpp/roaring.hh
@@ -691,6 +691,17 @@ class RoaringSetBitForwardIterator final {
         return orig;
     }
 
+    type_of_iterator& operator--() { // prefix --
+        roaring_previous_uint32_iterator(&i);
+        return *this;
+    }
+
+    type_of_iterator operator--(int) { // postfix --
+        RoaringSetBitForwardIterator orig(*this);
+        roaring_previous_uint32_iterator(&i);
+        return orig;
+    }
+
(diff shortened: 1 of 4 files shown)


lemire:
@Brian-Esch Can you have a look at the failing tests, especially travis? There seems to be problems.

lemire:
In particular, please take into account that x<<64 is undefined when x is a 64-bit int. You may need to do something like x<<(y&63) to abide by the specification of the C language.

Brian-Esch:
Pushed changes to fix these!

lemire:
Thanks. Let us see what CI thinks...

lemire:
We are in good shape!

PR #193, opened 2018-11-12
Have C++ defaults behave like stl defaults

Brian-Esch:
- Change `bool ra_init()` to `void ra_init()`, initializes array to have
  zero capacity. Does not `malloc` and cannot fail. ABI breaking change

- Have `Roaring` ctor use new `ra_init()`

- Make `Roaring` move ctor and move assignment `noexcept`. This is important
  for performance when holding them in stl containers (avoids copying)

- Delete dtor, copy ctor, move ctor, copy assignment operator and move
  assignment operator for `RoaringSetBitForwardIterator` and `Roaring64Map`
  since these objects don't manage any lifetimes manually. The compiler
  provides these by _default_ to us

- Add `static_assert`s that C++ classes are `noexcept` moveable in cpp
  unit test file

### DIFF
--- a/include/roaring/roaring_array.h
+++ b/include/roaring/roaring_array.h
@@ -51,9 +51,9 @@ roaring_array_t *ra_create(void);
 bool ra_init_with_capacity(roaring_array_t *new_ra, uint32_t cap);
 
 /**
- * Initialize with default capacity
+ * Initialize with zero capacity
  */
-bool ra_init(roaring_array_t *t);
+void ra_init(roaring_array_t *t);
 
 /**
  * Copies this roaring array, we assume that dest is not initialized
--- a/cpp/roaring.hh
+++ b/cpp/roaring.hh
@@ -20,10 +20,7 @@ class Roaring {
      * Create an empty bitmap
      */
     Roaring() {
-        bool is_ok = ra_init(&roaring.high_low_container);
-        if (!is_ok) {
-            throw std::runtime_error("failed memory alloc in constructor");
-        }
+        ra_init(&roaring.high_low_container);
         roaring.copy_on_write = false;
     }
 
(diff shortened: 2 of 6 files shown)


lemire:
See CI, we have...

>   c:\projects\croaring\tests\cpp_unit.cpp(24): error C2338: Expected Roaring64 to be no except move constructable [C:\projects\croaring\build\tests\cpp_unit.vcxproj]
713

include/roaring/roaring_array.h
```diff
@@ -51,9 +51,9 @@ roaring_array_t *ra_create(void);
 bool ra_init_with_capacity(roaring_array_t *new_ra, uint32_t cap);
 
 /**
- * Initialize with default capacity
+ * Initialize with zero capacity
  */
-bool ra_init(roaring_array_t *t);
+void ra_init(roaring_array_t *t);
```
madscientist: Wouldn't this be marked noexcept as well?

Brian-Esch: I'm not entirely sure what you are asking.. That is a C function and cannot be marked as noexcept

madscientist: Oops I got lost.  You're right.  GCC does have a nothrow attribute you can add to C functions but it's not necessary here.

Brian-Esch:
It appears that the move constructors on MSVC are not marked as `noexcept`. Here is a stack overflow https://stackoverflow.com/questions/47604029/move-constructors-of-stl-containers-in-msvc-2017-are-not-marked-as-noexcept

So the question is, do we want to conditionally do this static assert if we are on gcc or clang, or just remove it all together: https://github.com/RoaringBitmap/CRoaring/pull/193/files/9d0f97ae04ce4bfbed307ed358bbc225f7f088ab#diff-597bc881be4cdb305734afd4ac73dc2dR24

Brian-Esch:
@lemire 
The static assert is not necessary, I just thought it was a good way to test that the classes are noexcept moveable. I am going to remove the static assertion for `Roaring64Map` because the standard does not specify the noexcept requirements of a `std::map`'s move ctor. I am also going to remove them for `RoaringSetBitForwardIterator` and `Roaring64MapSetBitForwardIterator` because it does not appear that iterators need to satisfy `MoveConstructible`: https://en.cppreference.com/w/cpp/named_req/Iterator.

Unless there are objections I would like to leave the static assert for the `Roaring` class, as a way of future proofing the code, since we have defined noexcept for it's move constructor

lemire:
This now looks good to me. Congrats!

PR #195, opened 2018-12-12
Added aarch64 and armhf to travis workflow

andreigudkov:
* Reworked travis workflow so that now different
  configurations are tested simultaneously
* Added additional travis jobs to build and test
  roaring bitmap for aarch64 and armhf.
  Unit tests are run by using qemu.
* Introduced new ROARING_LINK_STATIC flag to produce
  statically linked binaries (tests and benchmarks)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,11 +2,33 @@ language: c
 sudo: true
 compiler:
   - clang
+dist: xenial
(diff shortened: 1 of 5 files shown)


lemire:
Looks good. Great work.

PR #198, opened 2019-01-11
Drop inline function declaration in public api

saulius:
Removes `inline` declaration in public API of `roaring_bitmap_add_range` and `roaring_bitmap_remove_range`. I was trying to make [croaring-rs](https://github.com/saulius/croaring-rs) work on Windows [as requested](https://github.com/saulius/croaring-rs/pull/42). These two functions seem to cause issues with MSVC linker: [see test output here](https://travis-ci.org/saulius/croaring-rs/builds/478230256). After I removed `inline` declaration it seems to build and work [just fine](https://travis-ci.org/saulius/croaring-rs/builds/478235806).

I understand the purpose of inlining here, but can we do without it? 

@lemire

### DIFF
--- a/include/roaring/roaring.h
+++ b/include/roaring/roaring.h
@@ -259,10 +259,7 @@ void roaring_bitmap_add_range_closed(roaring_bitmap_t *ra, uint32_t min, uint32_
 /**
  * Add all values in range [min, max)
  */
-inline void roaring_bitmap_add_range(roaring_bitmap_t *ra, uint64_t min, uint64_t max) {
-  if(max == min) return;
-  roaring_bitmap_add_range_closed(ra, (uint32_t)min, (uint32_t)(max - 1));
-}
+void roaring_bitmap_add_range(roaring_bitmap_t *ra, uint64_t min, uint64_t max);
 
 /**
  * Remove value x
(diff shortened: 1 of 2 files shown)


lemire:
Your fix was fine, but the larger issue was that the source file were missing an "extern inline" qualifier. Adding this qualifier should fix everything.

Can you pull https://github.com/RoaringBitmap/CRoaring/commit/9598ce81e88d5dc15ea03cdcfc911c084370ef39 

and see what happens?

rentenmark:
Hey @lemire thanks for making these changes https://github.com/RoaringBitmap/CRoaring/compare/6154187fa18728039d757e0f959b120a526e2877...c34aae0e174bc0d7cc9e747ec9249d9079cc144d do you think this warrants a version bump? It would be great to see this as a bump in https://github.com/lemire/CRoaringUnityBuild as well.

lemire:
Sure it does, but I'd prefer to get confirmation that it solves the original problem. I'd close this PR and make a new release.

lemire:
Can someone check that it solves the linkage problem?

saulius:
Hi @lemire,

Thank you for your changes. I can confirm the latest master builds correctly [on Windows as well as other platforms](https://travis-ci.org/saulius/croaring-rs/builds/478676297). As @rentenmark mentioned - it would be nice to get a new release cut.

Will close this merge request then!

lemire:
Version 0.2.59 released with the above fix.

Issue #74, opened 2017-01-15
Loading from memory-mapped files + fast operations over serialized bitmaps


aaiyer:
Loading from memory-mapped files - this feature exists in the Java Roaring Bitmap library

lemire:
Indeed.

lemire:
This has been implemented in Go: https://github.com/RoaringBitmap/roaring . One useful insight is that we can use copy-on-write so that the resulting bitmaps can be safely modified even when the backing data is immutable.

richardzhangrui:
@lemire Hi Lemire. Is this(fast operation over serialized bitmap) implemented in CRoaring now? Thanks.

PR #199, opened 2019-02-24
"Frozen" serialization format

andreigudkov:
Container data is stored in a format identical
to their C memory layout.
Instead of full deserialization, a lightweight readonly
view can be created atop of provided buffer.
This reduces amount of allocations and copying
and is suitable for creating readonly bitmaps from
memory mapped files.

* Added new functions:
  roaring_bitmap_frozen_size_in_bytes()
  roaring_bitmap_frozen_serialize()
  roaring_bitmap_frozen_view()

* copy_on_write was moved into more generic roaring_array_t.flags
  To modify copy_on_write behavior, new functions were added:
  roaring_bitmap_get_copy_on_write()
  roaring_bitmap_set_copy_on_write()

* roaring_bitmap_free() now takes pointer to a constant bitmap


lemire:
Fantastic. I consider that this is a satisfactory fix for https://github.com/RoaringBitmap/CRoaring/issues/74

PerpEternal:
@andreigudkov Hello, why

andreigudkov:
> @andreigudkov Hello, why roaring_bitmap_frozen_view input buf must align 32, would it still work if I remove this restriction?
> 
> https://github.com/RoaringBitmap/CRoaring/blob/33da04d7cf922c954228488e71b915a66ecff741/src/roaring.c#L3215

If I recall everything correctly, then:
1. All containers of the bitset type are located in the very beginning of the buffer
2. Bitset container operations use AVX2 operations with `load_mm256_load_si256` (not `loadu`). This requires that input is aligned by 32 bytes.

Probably things have changed since frozen format was introduced, and maybe this check is not required anymore.

lemire:
> Bitset container operations use AVX2 operations with load_mm256_load_si256 (not loadu). This requires that input is aligned by 32 bytes.

They do not, at least not currently.

lemire:
@PerpEternal It would be possible to revisit the alignment requirement, but someone needs to review the code and make sure that it is safe to lift it, and to run tests. If you are interested, don't hesitate. Ideally, we would want unit tests to make sure that it all works out.

andreigudkov:
@lemire Are you sure that you do not depend on the alignment? I can see that currently on x64 the alignment is either 32 or 64 (when I created frozen format, it was always 32).

https://github.com/RoaringBitmap/CRoaring/blob/8fc18acd32481810acc7166e3faa98be5e1b5675/src/containers/bitset.c#L75

lemire:
> Are you sure that you do not depend on the alignment?

I am sure of nothing. I only wrote that it could be revisited.

PR #260, opened 2020-10-31
Unify ra_copy() and ra_overwrite() 

AE1020:
Finding a problem in ra_overwrite() that had been worked around in
ra_copy() drew attention to how similar the two routines were (and
that a difference was only a bug).

The semantics were that ra_copy() assumed it was writing into
uninitialized memory while ra_overwrite() was willing to reuse the
existing capacity of a roaring_array_t if it was big enough.

This reframes the roaring_bitmap_copy() routine as initializing a
list of containers of the appropriate size (so it won't need a
realloc_array()) and then calling ra_overwrite().  ra_copy() is
then deleted.

### DIFF
--- a/src/roaring.c
+++ b/src/roaring.c
@@ -383,10 +383,16 @@ roaring_bitmap_t *roaring_bitmap_copy(const roaring_bitmap_t *r) {
     if (!ans) {
         return NULL;
     }
-    bool is_ok = ra_copy(&r->high_low_container, &ans->high_low_container,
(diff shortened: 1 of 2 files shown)


AE1020:
I had good intentions to try and pull an independent issue I looked at into a separate PR.  Unfortunately I realize now this requires the change to the C++ class here (which it was rebased out of):

https://github.com/RoaringBitmap/CRoaring/pull/259

PR #334, opened 2022-01-09
Cmocka download fetch content

Oppen:
Alternative version of #333 as suggested by @data-man.


Oppen:
Apparently drone's Ubuntu image and something about LGTM need to be updated to bump CMake.

Oppen:
@lemire do you know if there's any way to set the image used by LGTM.com? The one it's using now has cmake 3.13. The feature this PR uses is from 3.14.

PR #362, opened 2022-04-14
[Feature] implement `bitmap_contains_multi`

arthurkiller:
`roaring_bitmap_contains_multi` checks the bits by the given offset array and stores the value into results.


lemire:
That's reasonable. You are missing...

- [ ] tests!!!
- [ ] benchmarks (it should be faster in at least some cases compared to a naive implementation)
- [ ] header file version
- [ ] documentation (at least some)

src/roaring.c
```diff
@@ -129,6 +129,29 @@ void roaring_bitmap_add_many(roaring_bitmap_t *r, size_t n_args,
     }
 }
 
+void roaring_bitmap_contains_multi(roaring_bitmap_t *r, size_t n_args,
+                             const uint32_t *vals, bool *results) {
+    if (n_args == 0) return;
+    memset(results, 0, n_args);
```
Oppen, line 135: `sizeof(bool)` is not guaranteed to be `1`, so this may not clean as many results as we expect.
Besides, it may be best to just specify that `results` should be `0`-initialized and avoid duplicating computation when it is.

src/roaring.c
```diff
@@ -129,6 +129,29 @@ void roaring_bitmap_add_many(roaring_bitmap_t *r, size_t n_args,
     }
 }
 
+void roaring_bitmap_contains_multi(roaring_bitmap_t *r, size_t n_args,
+                             const uint32_t *vals, bool *results) {
+    if (n_args == 0) return;
```
Oppen, line 134: I'd rather have the `return` in its own line, some may miss it.

Oppen: And even then, the loop guard already makes sure nothing really interesting happens if we remove the `memset`. The fixed overhead of setting a few `0`s isn't likely to make a big difference.

src/roaring.c
```diff
@@ -129,6 +129,29 @@ void roaring_bitmap_add_many(roaring_bitmap_t *r, size_t n_args,
...
+            results[i] = container_contains(container, vals[i] & 0xFFFF, typecode);
+        } else {
+            containerindex = ra_get_index(&r->high_low_container, vals[i] >> 16);
+            if (containerindex >= 0) { // find the container
+                container = ra_get_container_at_index(&r->high_low_container, containerindex, &typecode);
+                // rest might be a tad expensive, possibly involving another round of binary search
+                results[i] = container_contains(container, vals[i] & 0xFFFF, typecode);
+            } // case else: just let results[i] false
+        }
+        prev = vals[i];
```
Oppen, line 151: I think this could be part of the `else` case.

src/roaring.c
```diff
@@ -129,6 +129,29 @@ void roaring_bitmap_add_many(roaring_bitmap_t *r, size_t n_args,
...
+    memset(results, 0, n_args);
+    uint32_t prev = 0;       // previous key
+    void *container = NULL;  // hold value of last container touched
+    uint8_t typecode = 0;    // typecode of last container touched
+    int containerindex = 0;
+    for (size_t i = 0; i < n_args; i++) {
+        if ((container != NULL) && (((prev ^ vals[i]) >> 16) == 0)) {
+            results[i] = container_contains(container, vals[i] & 0xFFFF, typecode);
+        } else {
+            containerindex = ra_get_index(&r->high_low_container, vals[i] >> 16);
```
Oppen, line 144: You can use `ra_advance_until` here to continue the search from where you left off.

Oppen: Note this part does not hold unless we assume sorted arguments.

Oppen: But even in that case, you could check if the new key is greater and it would then be valid.

src/roaring.c
arthurkiller: agreed

src/roaring.c
arthurkiller: fixed

src/roaring.c
arthurkiller: `prev` should be updated after each iteration, no matter which case we hit.

src/roaring.c
arthurkiller: > Note this part does not hold unless we assume sorted arguments.

That's what I also want to discuss. Should we add a high-bit sort here for the `vals` here? 

For most scenarios, users only have a few `vals` (maybe less than 2000) need to check if exist. And the cost of the sort (which may cause the memory copy such as `qsort`) may be larger than finding another `container`.

src/roaring.c
Oppen: We only look at the high bits in the checks. In the `if` case those are the same as for the previous value.

src/roaring.c
Oppen: I don't think that's necessary, at least not for now. In practice it will depend a lot on the kind of input. I don't think sorting it ourselves help a lot if we use a comparison sort that will be `O(nlogn)` anyway.
However, if we did check whether the next argument is greater than the previous one and did `ra_advance_until` in that case and full search otherwise we would be covering the case where the caller gives us ordered arguments, so we could at least advice on doing that in documentation.

src/roaring.c
Oppen: In any case it won't make a big difference, so do whatever you feel most comfortable with.

src/roaring.c
lemire: You do not have to change it, but I prefer...

```C++
if (n_args == 0) { return; }
```

src/roaring.c
lemire: At a minimum, I think we should go with...

```C++
for(...) { results[k] = false; }
```

if we are going to initialize at all. It does not seem warranted to use `memset`.

src/roaring.c
arthurkiller: I get it.

src/roaring.c
arthurkiller: > You can use `ra_advance_until` here to continue the search from where you left off.

I've tried this utility but it could add complexity to the logic. Once I want to run `ra_advance_until`, I need to check for 3 cases:
1. if the next value is larger than the previous one, if true we can call this with the last container index. Otherwise, the search will start from the very beginning (that means the input values are not sorted).
2. once the `ra_advance_until` is returned, I have to check the return value is equal to the container length. If true, next `ra_advance_until` should start from `0`.
3. If we got a good container position, still need to check the __current value__ contains in the `container` at the current position.

arthurkiller: Compare with using `ra_get_index` which means using `B-Search` instead of `Galloping search`, I think the benefits are very minuscule, and the extra complicity added could be unacceptable.

arthurkiller: But, anyway, I have implemented it already. LOL

Oppen: Looks good.

arthurkiller:
@Oppen 
Hi I have a question about `ra_advance_until(const roaring_array_t *ra, uint16_t x, int32_t pos)`,
it seems this will jump over the container position `pos`, which means if you want to search for `x` from the beginning to the end, you should consider pass __-1__ as the `pos`? Is that true?

arthurkiller:
I have new committed on this, please take a look

Oppen:
> Hi I have a question about `ra_advance_until(const roaring_array_t *ra, uint16_t x, int32_t pos)`,
> it seems this will jump over the container position `pos`, which means if you want to search for `x` from the beginning to the end, you should consider pass **-1** as the `pos`? Is that true?

That's correct.

src/roaring.c
```diff
@@ -129,6 +129,47 @@ void roaring_bitmap_add_many(roaring_bitmap_t *r, size_t n_args,
...
+    uint8_t typecode = 0;       // typecode of last container touched
+    int32_t container_index = -1;    // current container pos
+    const int container_count= r->high_low_container.size;
+
+    for (size_t i = 0; i < n_args; i++) {
+        if ((container != NULL) && (((prev_container_key^ vals[i]) >> 16) == 0)) {
+            // currently still in the same container
+            results[i] = container_contains(container, vals[i] & 0xFFFF, typecode);
+        } else {
+            if (container_index != -1 && vals[i] < (prev_container_key & 0xFFFF0000)) {
```
Oppen, line 148: Simply assigning every time the second condition is true is both simpler and probably more performant (branches are more expensive than assigning a constant in modern processors):
```suggestion
            if (vals[i] < (prev_container_key & 0xFFFF0000)) {
```

src/roaring.c
```diff
@@ -129,6 +129,47 @@ void roaring_bitmap_add_many(roaring_bitmap_t *r, size_t n_args,
...
+                // current value belong to the lower container, reset the index
```
Oppen, line 149:
```suggestion
                // current value belong to smaller key, reset the index
```

src/roaring.c
```diff
@@ -129,6 +129,47 @@ void roaring_bitmap_add_many(roaring_bitmap_t *r, size_t n_args,
...
+    uint32_t prev_container_key = 0;   // previous container key
+    void *container = NULL;     // hold value of last container touched
+    uint8_t typecode = 0;       // typecode of last container touched
+    int32_t container_index = -1;    // current container pos
+    const int container_count= r->high_low_container.size;
+
+    for (size_t i = 0; i < n_args; i++) {
+        if ((container != NULL) && (((prev_container_key^ vals[i]) >> 16) == 0)) {
+            // currently still in the same container
+            results[i] = container_contains(container, vals[i] & 0xFFFF, typecode);
```
Oppen, line 146:
```suggestion
            results[i] = container_contains(container, vals[i] & 0xFFFF, typecode);
            continue;
```

src/roaring.c
Oppen, line 148:
```suggestion
            if (vals[i] < (prev_container_key & 0xFFFF0000)) {
```

Oppen:
Looks mostly OK. We still need tests and documentation, and benchmarks would be desirable too.
@lemire what did you mean by header version BTW?

arthurkiller:
> Looks mostly OK. We still need tests and documentation, and benchmarks would be desirable too. @lemire what did you mean by header version BTW?

OK
I will add this later.

lemire:
We cannot merge without benchmarks and unit tests. The benchmarks are important because this function is pointless unless it can sometimes be faster than the point-by-point approach.

tests/toplevel_unit.c
```diff
@@ -57,6 +57,46 @@ DEFINE_TEST(range_contains) {
...
+
+    size_t length = 10;
+    bool results[length];
+    bool expected_results[length];
+    uint32_t values[length];
+    for (size_t i = 0; i < length; i+=2) {
+        expected_results[i] = 1;
+        expected_results[i+1] = 0;
+    }
+    values[0] =  1000;   // 1
```
Dr-Emann, line 80: We don't expect 1000 to be in the bitmap: `roaring_bitmap_from_range` inserts the range from `[start, end)` (exclusive of the top value).

tests/toplevel_unit.c
```diff
@@ -57,6 +57,46 @@ DEFINE_TEST(range_contains) {
...
+    values[1] =  1001;   // 0
+    values[2] =  66000;  // 1
+    values[3] =  66001;  // 0
+    values[3] =  66002;  // 1
```
Dr-Emann, line 84: This overwrites `values[3]`

tests/toplevel_unit.c
arthurkiller: that’s it, I will fix this later

arthurkiller:
benchmarking testing result on my MacBookPro
```
Data:
  cardinality: 1015367
Cycles/element:
  roaring_bitmap_contains: 20.309929 19.774353 49.094340 22.242483 20.091069
  roaring_bitmap_contains_multi: 19.149179 19.673742 38.943396 21.253152 20.512871
  roaring_bitmap_contains with sorted input: 15.373762 15.666473 39.509434 17.280310 15.216537
  roaring_bitmap_contains_multi with sorted input: 15.210349 15.532422 34.943396 16.548982 15.580181
```

arthurkiller:
seems CI did not works well on Arm arch

tests/toplevel_unit.c
```diff
@@ -57,6 +57,45 @@ DEFINE_TEST(range_contains) {
...
+    for (uint32_t i = 77000; i < 87000; i+=2) {
+        roaring_bitmap_add(bm, i);
+    }
+    // add bitset container from 20000
+    for (uint32_t i = 132000; i < 140000; i+=2) {
+        roaring_bitmap_add(bm, i);
+    }
+
+    size_t length = 10;
+    bool results[length];
```
lemire, line 73: This is not correct strictly speaking. 
Try:

```
const size_t length = 10;
bool results[length];
```

lemire:
> seems CI did not works well on Arm arch

It is not ARM per se. 

To declare an array, you must have a constant length.

tests/toplevel_unit.c
arthurkiller: fixed
My compiler did not even WARNING me about this.

lemire: I expect that @Dr-Emann is correct. In any case, the code needs fixing.

arthurkiller: > > Why it is not recommended to use stack memory?
> 
> It is _recommended_ to prefer stack when possible, but stack memory is limited... in some context it can be relatively scarce. If the stack memory is insufficient, you get an unsafe crash.
> 
> The following is definitively fine:
> 
> ```
> bool results[10];
> ```
> 
> Please don't do...
> 
> ```
> bool *results = (bool *)malloc(10 * sizeof(bool));
> ```
> 
> In the case where you do not know the size of the array (e.g., it could be 10, 1000, 1000000), then you cannot use stack allocation.


I get it, thanks a lot

Dr-Emann: It's not a question of suggested, or not suggested. In C without VLAs, you _cannot_ use the stack for an allocation which does not have a known size at compile time. `argc` is the count of the arguments passed to the executable: it _cannot_ be known until the code is running, and therefore _cannot_ be used as the size of an array.

Your options are:
* Pick an upper limit, and use that size:
    `if (argc > 100) abort(); uint32_t *values[100]; ....`
* Allocate a dynamic size using malloc:
    `uint32_t **values = malloc(argc * sizeof(*values));`

You can pass `-Werror=vla` to gcc to get errors locally to validate everything is correct (`cmake -DCMAKE_C_FLAGS="-Werror=vla" -DCMAKE_CXX_FLAGS="-Werror=vla" .`

@lemire if we want to support msvc, it might be worth adding `-Werror=vla` to the compile flags, so we get early errors when building locally outside of msvc?

include/roaring/roaring_array.h
```diff
@@ -95,6 +95,11 @@ inline container_t *ra_get_container_at_index(
  */
 uint16_t ra_get_key_at_index(const roaring_array_t *ra, uint16_t i);
 
+/**
+ * Check whether value should be placed in the container at index i
+ */
+inline bool ra_test_value_at_index(const roaring_array_t *ra, uint16_t i, uint32_t value);
```
Dr-Emann, line 101: This `inline`/`extern inline` is backwards (and should emit a warning: "inline function 'ra_test_value_at_index' is not defined". The definition should be here (otherwise, how can it inline?), and the `extern inline` in the `.c` file should be a declaration only:
```c
// In the header
inline bool ra_test_value_at_index(..) {
  // body here
}

// In the .c file, this will cause the single real instantiation of the function
extern inline bool ra_test_value_at_index(..);
```

Alternatively, just define the function as `static inline` in the header, and don't mention it in the `.c` file at all.

tests/toplevel_unit.c
lemire: > it might be worth adding

PR invited but note that it is not a huge problem in practice.

include/roaring/roaring_array.h
arthurkiller: that's reasonable, it should be `static inline`
`extern inline` makes me feel confused about how it would be compiled by the compiler, unfold or just marked as an outer symbol for calling.

arthurkiller: > This `inline`/`extern inline` is backwards (and locally for me emits a warning: "inline function 'ra_test_value_at_index' is not defined"). The definition should be here (otherwise, how can it inline?), and the `extern inline` in the `.c` file should be a declaration only:
> 
> ```c
> // In the header
> inline bool ra_test_value_at_index(..) {
>   // body here
> }
> 
> // In the .c file, this will cause the single real instantiation of the function
> extern inline bool ra_test_value_at_index(..);
> ```
> 
> Alternatively, just define the function as `static inline` in the header, and don't mention it in the `.c` file at all.

But you can find some `extern inline` in this repo... https://github.com/RoaringBitmap/CRoaring/issues/325

arthurkiller: @Dr-Emann I have questioned about this before.

include/roaring/roaring_array.h
```diff
@@ -95,6 +95,14 @@ inline container_t *ra_get_container_at_index(
  */
 uint16_t ra_get_key_at_index(const roaring_array_t *ra, uint16_t i);
 
+/**
+ * Check whether value should be placed in the container at index i
+ */
+static inline bool ra_test_value_at_index(const roaring_array_t *ra, uint16_t i, uint32_t value) {
```
lemire, line 101: Do we need this as a new function? It is a single line and it is not clear that it improves the quality of the code.

When someone is reading your code, they have to go look up this one function, and the function spans a single line.

And why is it coded this way? Don't you just want

```
ra_get_key_at_index(ra, i) == (uint16_t)(value >> 16)
```

Why the xor ?

include/roaring/roaring.h
```diff
@@ -335,6 +335,13 @@ bool roaring_bitmap_contains_range(const roaring_bitmap_t *r,
                                    uint64_t range_start,
                                    uint64_t range_end);
 
+/**
+ * Check multiple values sized by n_args return the existence of
+ * each element at the index by results.
+ */
```
lemire, line 341: You probably need to tell us something about the parameters. E.g., I suppose that results must have n_args elements?

src/roaring.c
```diff
@@ -129,6 +129,46 @@ void roaring_bitmap_add_many(roaring_bitmap_t *r, size_t n_args,
...
+            continue;
+        }
+
+        if (vals[i] < (prev_container_key & 0xFFFF0000)) {
+            // current value belong to the lower container, reset the index then search from beganing
```
lemire, line 151: beginning

src/roaring.c
```diff
@@ -129,6 +129,46 @@ void roaring_bitmap_add_many(roaring_bitmap_t *r, size_t n_args,
...
+        return;
+    }
+    uint32_t prev_container_key = 0;   // previous container key
+    void *container = NULL;     // hold value of last container touched
+    uint8_t typecode = 0;       // typecode of last container touched
+    int32_t container_index = -1;    // current container pos
+    const int container_count= r->high_low_container.size;
+
+    for (size_t i = 0; i < n_args; i++) {
+        if ((container != NULL) && (((prev_container_key^ vals[i]) >> 16) == 0)) {
```
lemire, line 144: `prev_container_key^ vals[i]` ---> `prev_container_key ^ vals[i]`

src/roaring.c
```diff
@@ -129,6 +129,46 @@ void roaring_bitmap_add_many(roaring_bitmap_t *r, size_t n_args,
     }
 }
 
+void roaring_bitmap_contains_multi(roaring_bitmap_t *r, size_t n_args,
+                             const uint32_t *vals, bool *results) {
+    if (n_args == 0 || r == NULL)  {
+        return;
+    }
+    uint32_t prev_container_key = 0;   // previous container key
```
lemire, line 137: I think that prev_container_key is just the previous key... nothing to do with containers, so could be `prev_val`, for example. (Simpler)

src/roaring.c
```diff
@@ -129,6 +129,46 @@ void roaring_bitmap_add_many(roaring_bitmap_t *r, size_t n_args,
...
+            container_index = -1;
+            continue;
+        }
+        if (ra_test_value_at_index(&r->high_low_container, container_index, vals[i])) {
+            container = ra_get_container_at_index(&r->high_low_container, container_index, &typecode);
+            results[i] = container_contains(container, vals[i]&0xFFFF, typecode);
+            prev_container_key = vals[i]; // update previous key, cause we really found a container
+        } else {
+            results[i] = false;
+            container_index = -1;
```
lemire, line 167: Why are you setting it to -1 here?

lemire:
@arthurkiller Presumably, you did run the benchmarking code that you wrote and contributed. Can you share your results?

lemire:
Please review the related PR https://github.com/RoaringBitmap/CRoaring/pull/363

src/roaring.c
arthurkiller: renew this key only when we went to another container.

include/roaring/roaring_array.h
arthurkiller: nothing special

fixed

src/roaring.c
arthurkiller: image
FYI

arthurkiller: image
FYI

arthurkiller:
> @arthurkiller Presumably, you did run the benchmarking code that you wrote and contributed. Can you share your results?

image

FYI

arthurkiller:
> Please review the related PR #363

It looks good to me, and I'm sure this would benefit us a lot.
I will do some benchmarks later.

arthurkiller:
> @arthurkiller Presumably, you did run the benchmarking code that you wrote and contributed. Can you share your results?



> benchmarking testing result on my MacBookPro
> 
> ```
> ./containsmulti_benchmark ../../benchmarks/realdata/weather_sept_85/*
> 
> Data:
>   cardinality: 1015367
> Cycles/element:
>   roaring_bitmap_contains: 20.309929 19.774353 49.094340 22.242483 20.091069
>   roaring_bitmap_contains_multi: 19.149179 19.673742 38.943396 21.253152 20.512871
>   roaring_bitmap_contains with sorted input: 15.373762 15.666473 39.509434 17.280310 15.216537
>   roaring_bitmap_contains_multi with sorted input: 15.210349 15.532422 34.943396 16.548982 15.580181
> ```

@lemire FYI

Dr-Emann:
I'm interested if we could do something similar to #363 for this PR, generalize by storing the intermediate state in a struct, which avoids requiring the caller put everything in a flat array, etc.

arthurkiller:
> 
You can check the benchmark result above. Sorted data can enhance the performance by about __20%__. But this result did not count the time `QSort` spent (even though we can sort the high 16 bits only, the memory movement is expensive).

But for me, adding a sort inside a lib API is not a good idea rather than prompting users to pass the sorted data instead if they want better performance. AFAIK roaring bitmap is usually used for _users tag selection_ which processes data from the streaming system. Users can do the sorting by themselves if they want better performance.

I suggest you add some benchmarks to show what we can get with sorting inside the API.

include/roaring/roaring.h
```diff
@@ -336,8 +336,8 @@ bool roaring_bitmap_contains_range(const roaring_bitmap_t *r,
                                    uint64_t range_end);
 
 /**
- * Check multiple values sized by n_args return the existence of
- * each element at the index by results.
+ * Check multiple values(vals) sized by n_args.
+ * Returns the results sized by n_args represent the existence of each value.
```
Oppen, line 340: I think something more along the lines of:
```
Checks presence of n_args values contained in array vals in bitmap r, returning the results in array results
```

src/roaring.c
```diff
@@ -129,6 +129,47 @@ void roaring_bitmap_add_many(roaring_bitmap_t *r, size_t n_args,
...
+            continue;
+        }
+        if (ra_get_key_at_index(&r->high_low_container, temp_container_index) == (uint16_t)(vals[i] >> 16)) {
+            container = ra_get_container_at_index(&r->high_low_container, temp_container_index, &typecode);
+            results[i] = container_contains(container, vals[i]&0xFFFF, typecode);
+            prev_container_key = vals[i]; // update previous key, cause we really found a container
+            container_index = temp_container_index; // update the current container index
+        } else {
+            results[i] = false;
+        }
```
Oppen, line 169: Take this as a mere suggestion, but I think it would be easier to read if you reverted the conditional so the body of the `if` doesn't need to be enclosed in a new indentation level, like:
```suggestion
        if (ra_get_key_at_index(&r->high_low_container, temp_container_index) != (uint16_t)(vals[i] >> 16)) {
            results[i] = false;
            continue;
        }
        container = ra_get_container_at_index(&r->high_low_container, temp_container_index, &typecode);
        results[i] = container_contains(container, vals[i]&0xFFFF, typecode);
        prev_container_key = vals[i]; // update previous key, cause we really found a container
        container_index = temp_container_index; // update the current container index
```
But as with all style, it's a little bit of bike shedding, so ignore it if you so prefer.

tests/toplevel_unit.c
```diff
@@ -57,6 +57,44 @@ DEFINE_TEST(range_contains) {
...
+    // create rbm with RLE conatainer form 0 1000
+    roaring_bitmap_t *bm = roaring_bitmap_from_range(0, 1001, 1);
+    // add array container from 66000
+    for (uint32_t i = 77000; i < 87000; i+=2) {
+        roaring_bitmap_add(bm, i);
+    }
+    // add bitset container from 20000
+    for (uint32_t i = 132000; i < 140000; i+=2) {
+        roaring_bitmap_add(bm, i);
+    }
```
Oppen, line 70: The comments contradict the code here.

tests/toplevel_unit.c
```diff
@@ -57,6 +57,44 @@ DEFINE_TEST(range_contains) {
...
+
+    bool results[10];
+    bool expected_results[10];
+    uint32_t values[10];
```
Oppen, line 74: You could just initialized `values` here with what you're going to test.

tests/toplevel_unit.c
```diff
@@ -57,6 +57,44 @@ DEFINE_TEST(range_contains) {
...
+    for (size_t i = 0; i < 10; i+=2) {
+        expected_results[i] = 1;
+        expected_results[i+1] = 0;
+    }
```
Oppen, line 78: You could make this less ad-hoc by simply storing the result of the regular `contains` function for `values[i]`.

tests/toplevel_unit.c
```diff
@@ -94,10 +132,6 @@ void can_copy_empty(bool copy_on_write) {
     roaring_bitmap_free(bm2);
 }
 
-
-
-
-
```
Oppen, line 100: These white space changes seem unrelated.

Oppen:
> But for me, adding a sort inside a lib API is not a good idea rather than prompting users to pass the sorted data instead if they want better performance. AFAIK roaring bitmap is usually used for users tag selection which processes data from the streaming system. Users can do the sorting by themselves if they want better performance.

Since arguably one of the best ways (in terms of efficiency) to sort the input here is to make half a radix sort, I'd say providing a helper that does exactly that in the library could be a good idea.
Regardless, for doing it inside the function I'd suggest splitting that logic into one that does the sort step and then call the existing one, for the simple reason that unconditionally modifying the input would make this option less useful: maybe the caller actually wanted to use the array as-is after calling it.

Dr-Emann:
@arthurkiller I'm not sure where sorting came in. I was talking about taking this from `roaring_bitmap_contains_multi(bitmap, count, values[], results[])` to something like `roaring_bitmap_contains_bulk(bitmap, &context, value) -> bool`, where context stores the values we currently have as local variables in `roaring_bitmap_contains_multi`. This would allow for all kinds of more flexibility in the callers, for example:

```c
contains_bulk_context_t context = {0};
// values generated lazily
// caller decides exit condition
while (1) {
  if !(roaring_bitmap_contains_bulk(bitmap, &context, next_value())) {
    break;
  }
}
```

arthurkiller:
> 
Sorry for the misunderstood. Add a context looks reasonable.

arthurkiller:
> @arthurkiller I'm not sure where sorting came in. I was talking about taking this from `roaring_bitmap_contains_multi(bitmap, count, values[], results[])` to something like `roaring_bitmap_contains_bulk(bitmap, &context, value) -> bool`, where context stores the values we currently have as local variables in `roaring_bitmap_contains_multi`. This would allow for all kinds of more flexibility in the callers, for example:
> 
> ```c
> contains_bulk_context_t context = {0};
> // values generated lazily
> // caller decides exit condition
> while (1) {
>   if !(roaring_bitmap_contains_bulk(bitmap, &context, next_value())) {
>     break;
>   }
> }
> ```


Sorry for the misunderstood. Add a context looks reasonable.

arthurkiller:
> 
Should the `contains_bulk_context_t ` be a public container-seeking facility for `contains_multi` to access?

arthurkiller:
> @arthurkiller I'm not sure where sorting came in. I was talking about taking this from `roaring_bitmap_contains_multi(bitmap, count, values[], results[])` to something like `roaring_bitmap_contains_bulk(bitmap, &context, value) -> bool`, where context stores the values we currently have as local variables in `roaring_bitmap_contains_multi`. This would allow for all kinds of more flexibility in the callers, for example:
> 
> ```c
> contains_bulk_context_t context = {0};
> // values generated lazily
> // caller decides exit condition
> while (1) {
>   if !(roaring_bitmap_contains_bulk(bitmap, &context, next_value())) {
>     break;
>   }
> }
> ```


Should the contains_bulk_context_t be a public container-seeking facility for contains_multi to access?

lemire:
@arthurkiller Can you comment on @Oppen's review points?

include/roaring/roaring.h
```diff
@@ -335,6 +335,13 @@ bool roaring_bitmap_contains_range(const roaring_bitmap_t *r,
                                    uint64_t range_start,
                                    uint64_t range_end);
 
+/**
+ * Check multiple values(vals) sized by n_args.
+ * Returns the results sized by n_args represent the existence of each value.
+ */
+void roaring_bitmap_contains_multi(roaring_bitmap_t *r, size_t n_args,
```
Dr-Emann, line 342: I think this should take `const roaring_bitmap_t *`, since it shouldn't modify the bitmap.

Dr-Emann:
Yeah, I think both `add_bulk` and `contains_bulk` could use a common "context" struct, which stores the last accessed container, typecode, index, and key

arthurkiller:
@lemire @Oppen Shall we push forward this issue?
what's the word?

lemire:
@arthurkiller We are extending the API, and it would be nice to do it in a consistent fashion.

Both `add_bulk` and `contains_bulk` could use a common "context" structure, or a similar one at least. Possibly, we might want to take both PR and create just one PR.

I am nervous about extending the API in an ad hoc manner and releasing it... then it is hard to change. So we want to get the public-facing stuff right on the first release. Here, I don't think it is quite right.

arthurkiller:
> @arthurkiller We are extending the API, and it would be nice to do it in a consistent fashion.
> 
> Both `add_bulk` and `contains_bulk` could use a common "context" structure, or a similar one at least. Possibly, we might want to take both PR and create just one PR.
> 
> I am nervous about extending the API in an ad hoc manner and releasing it... then it is hard to change. So we want to get the public-facing stuff right on the first release. Here, I don't think it is quite right.

Such a pity, this facility has already been pushed to the online environment and really helped a lot on data loading scenery.
No matter how these facilities are drafted on this repo, I hope these APIs can help more people use CRoaring while loading massive data.

Closed via no more response this PR.

lemire:
@arthurkiller I think we do not understand each other. We did not reject this PR. In fact, we want to merge it. But we want that the public API be nice before we merge, so that we are not stuck later with an inconsistent API that we can't fix without breaking people's code.

lemire:
We will attempt to move forward. We do value this contribution, we just want to make sure it is as good as possible when we merge and make it public.

arthurkiller:
I’m busy these days, reply may be late.

Have an awesome day

✉️✉️✉️✉️✉️✉️✉️✉️✉️✉️✉️✉️
Arthur lee
Sent from my iPhone

PR #470, opened 2023-05-05
add get_idx function to return the 1 index of x

bryceustc:
When using the CRoaring library, the normal way to get the 1 index of a value in a RoaringBitmap is:
1. Call contains() to check if the value exists
2. If it exists, call rank() to get the 1 index
3. If it not exists, do something else
However, this requires traversing the RoaringBitmap twice, which is inefficient. To improve this, a get_idx() interface was created. It works as follows:
1. Traverse the RoaringBitmap once
2. If the value is found, return its 1 index in the set
4. If the value is not found, return -1

### DIFF
--- a/cpp/roaring.hh
+++ b/cpp/roaring.hh
@@ -500,6 +500,17 @@ public:
     }
 
     /**
+     * Returns the 1 index of x in the bitmap.
+     * If the bitmap dosen't contain x , this function will return -1.
+     * The difference with rank function is that this function will return -1
+     * when x isn't in the bitmap, but the rank funciton will return a
+     * non-negative number.
+     */
+    int64_t get_idx(uint32_t x) const noexcept {
+        return api::roaring_bitmap_get_idx(&roaring, x);
+    }
+
+    /**
      * Write a bitmap to a char buffer. This is meant to be compatible with
      * the Java and Go versions. Returns how many bytes were written which
      * should be getSizeInBytes().
--- a/src/roaring.c
+++ b/src/roaring.c
@@ -2678,6 +2678,34 @@ uint64_t roaring_bitmap_rank(const roaring_bitmap_t *bm, uint32_t x) {
 }
 
 /**
+ * roaring_bitmap_get_idx returns the 1 idx of x, if not exsist return -1.
+ */
+int64_t roaring_bitmap_get_idx(const roaring_bitmap_t *bm, uint32_t x) {
+    int64_t index = 0;
+    const uint16_t xhigh = x >> 16;
+    int32_t high_idx = ra_get_index(&bm->high_low_container, xhigh);
+    if (high_idx < 0) return -1;
+
+    for (int i = 0; i < bm->high_low_container.size; i++) {
+        uint32_t key = bm->high_low_container.keys[i];
+        if (xhigh > key) {
+            index +=
+                container_get_cardinality(bm->high_low_container.containers[i],
+                                          bm->high_low_container.typecodes[i]);
+        } else if (xhigh == key) {
+            int32_t low_idx = container_get_idx(
+                bm->high_low_container.containers[high_idx],
+                bm->high_low_container.typecodes[high_idx], x & 0xFFFF);
+            if (low_idx < 0) return -1;
+            return index + low_idx;
+        } else {
+            return -1;
+        }
+    }
+    return index;
+}
+
+/**
 * roaring_bitmap_smallest returns the smallest value in the set.
 * Returns UINT32_MAX if the set is empty.
 */
--- a/tests/toplevel_unit.c
+++ b/tests/toplevel_unit.c
@@ -3512,6 +3512,23 @@ static uint64_t rank(uint32_t *arr, size_t length, uint32_t x) {
     return sum;
 }
 
+static int64_t get_idx(uint32_t *arr, size_t length, uint32_t x) {
+    bool is_present = false;
+    for (size_t i = 0; i < length; ++i) {
+        if (arr[i] == x) {
+            is_present = true;
+        }
+    }
+    if (!is_present) return -1;
+
+    uint64_t sum = 0;
+    for (size_t i = 0; i < length; ++i) {
+        if (arr[i] > x) break;
+        sum++;
+    }
+    return sum;
+}
+
 DEFINE_TEST(test_rank) {
     for (uint32_t mymin = 123; mymin < 1000000; mymin *= 2) {
         // just arrays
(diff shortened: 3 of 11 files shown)


lemire:
This is reasonable and potentially useful but while `contains` and `rank` are well known terms (to me at least), I am unfamiliar with the terminology `1 index`. Do you have a reference (maybe in other software) regarding this terminology?

If it is a terminology you created, we will have to discuss it further.

bryceustc:
Apologies for any confusion around the term '1 index'. My intent is to refer simply to the index of a given value in the RoaringBitmap. As an example, if a Roaring Bitmap is {2, 4, 5, 7, 8, 9, 10}, I want get the index of value 7, it will return 4, which means that 7 is the 4th 1 in the set.  If i want get the index of value 6, it will return -1, which means that 6 is not a element of the set. In a word, The proposed get_idx() interface aims to find the index (or -1 if not exsist) in a single traversal.

lemire:
@bryceustc  I do understand the concept.

I am concerned about the terminology.



I don't have much to propose, but "get_idx" and "1 index" won't work. You know what it means, but the average programmer won't be able to tell.

What about "get_index" and just say that we are return the index of the element?

bryceustc:
It's a good suggestion, I will modify the relevant code and descriptions. Thanks :)

bryceustc:
I have modified the relevant content according to your suggestions. Thank you in advance for reviewing again.

cpp/roaring.hh
```diff
@@ -499,6 +499,17 @@ public:
...
     }
 
+    /**
+     * Returns the index of x in the set.
+     * If the set dosen't contain x , this function will return -1.
+     * The difference with rank function is that this function will return -1
+     * when x isn't in the set, but the rank funciton will return a
+     * non-negative number.
+     */
+    int64_t get_index(uint32_t x) const noexcept {
```
longqimin, line 509: @bryceustc nice work. but the name may be a little confusing. `index` start from 0 in general, while the `rank` value start from 1

bryceustc: 👍 Thanks for your comments, it should start from 0. I have modified the relevant codes and descriptions, please review again when convenient :)

tests/toplevel_unit.c
```diff
@@ -3512,6 +3512,19 @@ static uint64_t rank(uint32_t *arr, size_t length, uint32_t x) {
     return sum;
 }
 
+static int64_t get_index(uint32_t *arr, size_t length, uint32_t x) {
+    bool is_present = false;
+    size_t i = 0;
+    for (; i < length; ++i) {
+        if (arr[i] == x) {
+            is_present = true;
```
longqimin, line 3520: you can `return i;` here for simplicity

bryceustc: Thank you for your meticulous work!

Dr-Emann:
Maybe it would be worthwhile to take a page from binarySearch, and return `-(i+1)` rather than always returning -1, so you can know what the index of the number would be if it _were_ in the bitmap.

`roaring_bitmap_select` is the opposite of this function, we should mention it in this function, and vice versa.

cpp/roaring.hh
```diff
@@ -499,6 +499,17 @@ public:
...
     }
 
+    /**
+     * Returns the index of x in the set, index start from 0.
+     * If the set dosen't contain x , this function will return -1.
+     * The difference with rank function is that this function will return -1
+     * when x isn't in the set, but the rank funciton will return a
+     * non-negative number.
+     */
+    int64_t get_index(uint32_t x) const noexcept {
```
Dr-Emann, line 509: The other C++ methods are `camelCase`, I think this should be `getIndex`

Do we want to add the same to `roaring64map.hh`?

Dr-Emann (commented):
About naming, not many of the roaring functions have `get` in them, even when they're logically getters. I wonder if a better name might be `roaring_bitmap_index_of`, or something referencing `select` since that's the dual of this function, like `roaring_bitmap_selection_index`?

src/roaring.c
```diff
@@ -2677,6 +2677,34 @@ uint64_t roaring_bitmap_rank(const roaring_bitmap_t *bm, uint32_t x) {
...
+        if (xhigh > key) {
+            index +=
+                container_get_cardinality(bm->high_low_container.containers[i],
+                                          bm->high_low_container.typecodes[i]);
+        } else if (xhigh == key) {
+            int32_t low_idx = container_get_index(
+                bm->high_low_container.containers[high_idx],
+                bm->high_low_container.typecodes[high_idx], x & 0xFFFF);
+            if (low_idx < 0) return -1;
+            return index + low_idx;
```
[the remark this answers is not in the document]
Dr-Emann: Ah, never mind, it would take a call to both `rank` and `select`, and either would have to do some extra work in some cases.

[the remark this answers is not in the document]
Dr-Emann: I think it's probably possible to implement both `rank` and `get_index` using a common function for containers, if the common function does the `return the negative value of where it would be`

cpp/roaring64map.hh
```diff
@@ -1092,6 +1093,23 @@ public:
...
+     */
+    int128_t getIndex(uint64_t x) const {
+        uint64_t index = 0;
+        auto roaring_destination = roarings.find(highBytes(x));
+        if (roaring_destination != roarings.cend()) {
+            for (auto roaring_iter = roarings.cbegin();
+                 roaring_iter != roaring_destination; ++roaring_iter) {
+                index += roaring_iter->second.cardinality();
+            }
+            index += roaring_destination->second.getIndex(lowBytes(x));
```
longqimin, line 1107: @bryceustc  it is a bug to add `-1` to `index`

longqimin: and the type of `index` should be `int128_t`

cpp/roaring64map.hh
```diff
@@ -1092,6 +1093,23 @@ public:
         return result;
     }
 
+    /**
+     * Returns the number of integers that are smaller or equal to x.
```
longqimin, line 1097: the comments is misleading

cpp/roaring64map.hh
bryceustc: Thanks for your review, the bug and comments have been fixed in the latest commit.

cpp/roaring64map.hh
```diff
@@ -1092,6 +1093,29 @@ public:
...
     }
 
+    /**
+     * Returns the index of x in the set, index start from 0.
+     * If the set doesn't contain x , this function will return -1.
+     * The difference with rank function is that this function will return -1
+     * when x isn't in the set, but the rank function will return a
+     * non-negative number.
+     */
+    int128_t getIndex(uint64_t x) const {
```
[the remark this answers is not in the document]
bryceustc: Sorry, I'm afraid I didn't grasp what you meant...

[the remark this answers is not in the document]
Dr-Emann: I don't think we need to use int128_t, because if you ever had an item at an index greater than an int64_t can represent, you're already out of memory. The `rank` method returns a uint64_t, and I argue this method should return an int64_t.

The most efficient way to get there is to have a range container for every slot in every bitmap. Each range container can cover 0x10000 items. So to get to 0x8000_0000_0000_0000 (the first index which can't be returned in an i64), you need 0x0000_8000_0000_000 range containers (140737488355328). If each range container (and the keys array, and the container type array, and malloc overhead, and...) magically took just 1 byte per container (spoiler, it will take a LOT more than that), you'd still need 128TiB of memory. There's simply no way to ever realistically have a RoaringBitmap64 with more than 2^63 items.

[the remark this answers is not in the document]
lemire: My math is slightly more pessimistic. Each container can represent 65536 values. I get 2**64/(2**16) is  281474976710656. Assuming one byte per container that is... 256 TiB.

I think that Windows cannot ever support more than 6 TB of memory... https://learn.microsoft.com/fr-ca/windows/win32/memory/memory-limits-for-windows-releases?redirectedfrom=MSDN

A reasonable solution here is to use a 64-bit value and explain why you do so.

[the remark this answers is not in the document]
Dr-Emann: Signed integer means `2^63` rather than `2^64`. I made that mistake initially, you can see I said 256 TB in my first message. Not that it matters, either way, even assuming an impossible 1 byte per container, it's not gonna fit in even virtual memory.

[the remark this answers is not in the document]
lemire: Right ok. Most systems have a theoretical limit of maybe 2^56 bytes per process. Of course, this may change in the future, but before that happens, int128_t will be part of the every standard programming language.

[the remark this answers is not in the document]
bryceustc: 👍 Thank you all for the professional and detailed explanations. I have changed `int128_t` to `int64_t` in the latest commit.

lemire:
Running tests.

lemire:
@Dr-Emann @longqimin This PR looks good to me.

Any objection to merging?

longqimin:
> @Dr-Emann @longqimin This PR looks good to me.
> 
> Any objection to merging?

LGTM too

PR #493, opened 2023-06-19
Add a function to do internal validations of the bitmap, intended for testing

Dr-Emann:
Adds the function `roaring_bitmap_internal_validate` which validates assumptions that other code relies on (e.g. arrays are sorted and non-duplicated).

This may be useful for e.g. validating the result of deserializing, but mostly, it allows for an easy check to sprinkle through the unit tests to ensure the bitmap is always in a consistent state.


lemire (approved):
Very good.

PR #548, opened 2024-01-13
Use >= instead of > for art_node*_prev_child

SLieve:
Using > caused it to skip the first child. Only applies to `node48` and `node256` Modified an existing test to cover this case.


Dr-Emann (approved):
Makes sense to me

PR #550, opened 2024-01-13
roaring64: container_remove_range returns null rather than returning an empty container

Dr-Emann:
This would lead to all sorts of issues: `typecode2` is read uninitialized, and hopefully just asserts that it's not one of the known typecodes. Then, we've stored a null container into the bitmap.


SLieve (approved):
Thanks for finding it and for the fix! I saw in https://github.com/RoaringBitmap/croaring-rs/pull/125 that you found most of these bugs with fuzzing, that's really cool :)

Dr-Emann:
The CI failure seems like it shouldn't be caused by my change, going to rebase to rerun the CI.

PR #551, opened 2024-01-13
roaring64: container_add_range _is_ inclusive

Dr-Emann:
`container_add_range` is from `[min, max]`, no need to add one: https://github.com/RoaringBitmap/CRoaring/blob/f3ad2f2e52c53d0241a6327bdf193423c47059c0/include/roaring/containers/containers.h#L2377

Add a test to verify adding a range to an existing container works correctly


SLieve (approved):
I see where I went wrong. Looks good, thanks!

PR #554, opened 2024-01-14
fix(roaring64): add_bulk of existing values no longer reads OOB

Dr-Emann:
When the bulk context was null, it assumed there was no leaf at all, but that's not nessisarily true.


SLieve (approved):
Makes sense, thanks!

PR #557, opened 2024-01-14
Use assert_int_equal instead of assert_true for integer comparisons

SLieve:
This should provide better error messages. Apparently `assert_int_equal` also allows comparisons of wider types like `uint64`.


Dr-Emann (approved):
Makes sense to me

PR #563, opened 2024-01-16
ART: Fix a few issues around lower bound / erase

### LINKED ISSUE #559: roaring64: and_inplace gives incorrect value
This is with all outstanding roaring64 PRs included.

Reproduction:
```c
uint64_t start = 0x0FFFF;
uint64_t end = 0x20001;

roaring64_bitmap_t* r1 = roaring64_bitmap_from_range(start, end, 1);
roaring64_bitmap_add(r1, 0xFFFF0000);

roaring64_bitmap_t* r2 = roaring64_bitmap_from_range(start, end, 1);

uint64_t and_cardinality = roaring64_bitmap_and_cardinality(r1, r2);
assert_int_equal(and_cardinality, end - start);

roaring64_bitmap_t* r3 = roaring64_bitmap_and(r1, r2);
assert_int_equal(roaring64_bitmap_get_cardinality(r3), and_cardinality);

roaring64_bitmap_and_inplace(r1, r2);
assert_int_equal(roaring64_bitmap_get_cardinality(r1), and_cardinality); // This line fails, `0x10001 != 0x10002`
assert_true(roaring64_bitmap_equals(r1, r3));

roaring64_bitmap_free(r1);
roaring64_bitmap_free(r2);
roaring64_bitmap_free(r3);
```

SLieve:
Fixes #559 

1. Leaf keys were written to the wrong location: `memcpy(&leaf->key` instead of `memcpy(leaf->key`
2. In `art_node_iterator_lower_bound`, we did not check if the key chunk for the lower bound child was greater than the key chunk we searched for.
3. In `art_iterator_erase` there were a few issues that would lead to an incorrect location. I've somewhat sidestepped the most tricky bit by doing a lower bound search from the root instead.

Added tests for 1 and 2. Also slightly improved `art_printf` for debugging.

### DIFF
--- a/src/art/art.c
+++ b/src/art/art.c
@@ -1178,7 +1178,15 @@ static void art_node_print_type(const art_node_t *node) {
 }
 
 void art_node_printf(const art_node_t *node, uint8_t depth) {
-    printf("%*s", depth, "");
+    if (art_is_leaf(node)) {
+        printf("{ type: Leaf, key: ");
+        art_leaf_t *leaf = CAST_LEAF(node);
+        for (size_t i = 0; i < ART_KEY_BYTES; ++i) {
+            printf("%02x", (char)(leaf->key[i]));
+        }
+        printf(" }\n");
+        return;
+    }
     printf("{\n");
     depth++;
 
@@ -1187,19 +1195,6 @@ void art_node_printf(const art_node_t *node, uint8_t depth) {
     art_node_print_type(node);
     printf("\n");
 
-    if (art_is_leaf(node)) {
-        art_leaf_t *leaf = CAST_LEAF(node);
-        printf("%*s", depth, "");
-        printf("key: ");
-        for (size_t i = 0; i < ART_KEY_BYTES; ++i) {
-            printf("%x", leaf->key[i]);
-        }
-        printf("\n");
-        depth--;
-        printf("%*s", depth, "");
-        printf("}\n");
-        return;
-    }
     art_inner_node_t *inner_node = (art_inner_node_t *)node;
     printf("%*s", depth, "");
     printf("prefix_size: %d\n", inner_node->prefix_size);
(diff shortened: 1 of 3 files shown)


Dr-Emann:
Think the change to `art_node_iterator_lower_bound` made for a possible infinite loop:

```c
roaring64_bitmap_t* r1 = roaring64_bitmap_create();
roaring64_bitmap_add(r1, 829185);
roaring64_bitmap_t* r2 = roaring64_bitmap_create();
roaring64_bitmap_add(r2, 0);
roaring64_bitmap_add(r2, 14418687);

roaring64_bitmap_and_cardinality(r1, r2); // This function never returns

roaring64_bitmap_free(r1);
roaring64_bitmap_free(r2);
```

It appears this function call doesn't actually move the iterator at all in this case: https://github.com/RoaringBitmap/CRoaring/blob/4a7474e81351ece3c3be80445793746997a0dded/src/roaring64.c#L823

SLieve:
Should be fixed with my latest commit. I aim to improve the test coverage to catch issues like these in the future.

Dr-Emann (approved):

Dr-Emann (changes requested):
```c
roaring64_bitmap_t *r1 = roaring64_bitmap_of(2, 0x000001, 0xFFFFF0001);
roaring64_bitmap_t *r2 = roaring64_bitmap_of(2, 0x100001, 0xFFFFF0001);

uint64_t card = roaring64_bitmap_and_cardinality(r1, r2);
assert_int_equal(card, 1); // This assert fails: `0 != 1`

roaring64_bitmap_free(r1);
roaring64_bitmap_free(r2);
```

SLieve:
Should be fixed with the latest commit.

Dr-Emann (approved):

PR #565, opened 2024-01-20
node48 children can have holes

Dr-Emann:
When removing from a node48, the child is simply set to NULL. For example, if the state is (ignoring that it actually needs at least 17 items to not be a node16):

```
keys: {[0]=0, [1]=1}
children: {[0]=child1, [1]=child2}
count: 2
```

After removing key 0, the state will be:

```
keys: {[0]=48, [1]=1}
children: {[0]=NULL, [1]=child2}
count: 1
```

If we then try to insert at key 3, the item at `children[count]` is already in use: we have to find and reuse the 0th child.

Rather than matching the [java implementation][1] (though see the first commit [7dbad58](https://github.com/RoaringBitmap/CRoaring/pull/565/commits/7dbad58b2e92654465f00c9fc92baa902f57e823)), keep a bitset of available children slots. This slightly increases the size of the node48 struct, but it means:

* We don't have to initialize all children to null at init time
* We don't have to loop looking for a child (sometimes) on insert

[1]: https://github.com/RoaringBitmap/RoaringBitmap/blob/1ed0242d1f298cf93d3666a754eb4bb4c5180c4c/RoaringBitmap/src/main/java/org/roaringbitmap/art/Node48.java#L181-L187

This also has some fixes for `art_printf`:

* When printing keys, don't cast to char, if char is signed, when integer promoted, a 0xFF char will be printed as `FFFFFFFF`
* Correctly output keys for node48
* Include child index for node48
* Correct child output for node48

### DIFF
--- a/src/art/art.c
+++ b/src/art/art.c
@@ -462,6 +462,9 @@ static art_node48_t *art_node48_create(const art_key_chunk_t prefix[],
     for (size_t i = 0; i < 256; ++i) {
         node->keys[i] = ART_NODE48_EMPTY_VAL;
     }
+    for (size_t i = 0; i < 48; ++i) {
+        node->children[i] = NULL;
+    }
     return node;
 }
 
@@ -488,6 +491,13 @@ static art_node_t *art_node48_insert(art_node48_t *node, art_node_t *child,
                                      uint8_t key) {
     if (node->count < 48) {
         uint8_t val_idx = node->count;
+        if (node->children[val_idx] != NULL) {
+            val_idx = 0;
+            // Find an empty child.
+            while (node->children[val_idx] != NULL) {
+                val_idx++;
+            }
+        }
         node->keys[key] = val_idx;
         node->children[val_idx] = child;
         node->count++;
@@ -1182,7 +1192,7 @@ void art_node_printf(const art_node_t *node, uint8_t depth) {
         printf("{ type: Leaf, key: ");
         art_leaf_t *leaf = CAST_LEAF(node);
         for (size_t i = 0; i < ART_KEY_BYTES; ++i) {
-            printf("%02x", (char)(leaf->key[i]));
+            printf("%02x", leaf->key[i]);
         }
         printf(" }\n");
         return;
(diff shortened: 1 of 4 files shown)


src/art/art.c
```diff
@@ -74,6 +77,8 @@ typedef struct art_node16_s {
 typedef struct art_node48_s {
     art_inner_node_t base;
     uint8_t count;
+    // Bitset where the ith bit is set if children[i] is available
+    uint64_t available_children;
```
SLieve, line 81: Now that we have this mask, we could use `64 - popcount` to determine the number of children, which would remove the need to keep `count` around.

src/art/art.c
```diff
@@ -74,6 +77,8 @@ typedef struct art_node16_s {
 typedef struct art_node48_s {
     art_inner_node_t base;
     uint8_t count;
+    // Bitset where the ith bit is set if children[i] is available
```
SLieve, line 80: Nit: I'd suggest mentioning that only the first 48 bits are used.

tests/roaring64_unit.cpp
```diff
@@ -952,6 +952,27 @@ DEFINE_TEST(test_andnot_inplace) {
     }
 }
 
+DEFINE_TEST(test_shrink_grow_48_art_node) {
```
SLieve, line 955: Since this is purely testing ART, what about moving this test into `art_unit.cpp`?

SLieve (commented):
Nice, thanks for finding and fixing!

src/art/art.c
Dr-Emann: I thought about doing this, but it doesn't save any size because of padding (even if we tried `uint8_t available_children[6]` to use 48 bits exactly), and I expect it's faster to keep the count separately, but I haven't benchmarked.

tests/art_unit.cpp
```diff
@@ -39,8 +38,7 @@ void assert_key_eq(art_key_chunk_t* key1, art_key_chunk_t* key2) {
 class Key {
    public:
     Key(uint64_t key) {
-        // Reverse byte order of the low 6 bytes. Not portable to big-endian
-        // systems!
+        // Store the low 6 bytes of the key in big-endian order.
```
Dr-Emann, line 41: This is actually totally portable

SLieve: Yes I think I used a different way to do this before and forgot to update the comment.

SLieve (approved):

Dr-Emann:
Forgot to free at the end of the test

Dr-Emann:
As a bonus, with this and the other PRs merged so far, I've now been fuzzing for a few hours with no issues 🎉

SLieve:
Nice!

PR #569, opened 2024-01-23
Implement roaring64_bitmap_to_uint64_array

SLieve:
Also added a benchmark.

### DIFF
--- a/src/roaring64.c
+++ b/src/roaring64.c
@@ -1395,18 +1395,11 @@ bool roaring64_bitmap_iterate(const roaring64_bitmap_t *r,
     return true;
 }
 
-static inline bool roaring64_bitmap_to_uint64_array_iterator(uint64_t value,
-                                                             void *param) {
-    uint64_t **array = (uint64_t **)param;
-    **(array) = value;
-    (*(array))++;
-    return true;
-}
-
 void roaring64_bitmap_to_uint64_array(const roaring64_bitmap_t *r,
                                       uint64_t *out) {
-    roaring64_bitmap_iterate(r, &roaring64_bitmap_to_uint64_array_iterator,
-                             (void *)(&out));
+    roaring64_iterator_t it = {0};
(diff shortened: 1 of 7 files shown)


Dr-Emann (approved):

src/roaring64.c
```diff
@@ -1395,6 +1395,20 @@ bool roaring64_bitmap_iterate(const roaring64_bitmap_t *r,
...
+    uint64_t **array = (uint64_t **)param;
+    **(array) = value;
+    (*(array))++;
+    return true;
+}
+
+void roaring64_bitmap_to_uint64_array(const roaring64_bitmap_t *r,
+                                      uint64_t *out) {
+    roaring64_bitmap_iterate(r, &roaring64_bitmap_to_uint64_array_iterator,
+                             (void *)(&out));
```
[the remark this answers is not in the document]
SLieve: Running the microbenchmark on my machine (a VM, take results with a grain of salt) it seems to be about 4x faster, so I've switched to that implementation.

Good catch on the count truncation bug!

src/roaring64.c
```diff
@@ -1395,6 +1395,13 @@ bool roaring64_bitmap_iterate(const roaring64_bitmap_t *r,
     return true;
 }
 
+void roaring64_bitmap_to_uint64_array(const roaring64_bitmap_t *r,
+                                      uint64_t *out) {
+    roaring64_iterator_t it = {0};
+    roaring64_iterator_init_at(r, &it, out);
```
Dr-Emann, line 1401: `roaring64_iterator_init_at`'s last parameter is `bool first`, I think we just want to always pass true here, not `out` (though presumably, `out` will always be non-null, though it may be valid to call this with null if the bitmap is empty)

SLieve: Facepalm. Thanks!

Dr-Emann (approved):

PR #578, opened 2024-01-29
r64 iterator: allow moving from beyond the start / end

### LINKED ISSUE #574: Roaring64 iterator cannot come back from the edge
Something like this should work

```c
roaring64_bitmap_t *b = roaring64_bitmap_of(1, 100ULL);
roaring64_iterator_t *it = roaring64_iterator_create(b);
// it points at the first item, 100

roaring64_iterator_previous(it);
// it now points before the first item

roaring64_iterator_advance(it);
assert(roaring64_iterator_has_value(it));
assert(roaring64_iterator_value(it) == 100);
```

But instead it fails, right at the beginning of `roaring64_iterator_advance`:

```c
if (it->art_it.value == NULL) {
    return (it->has_value = false);
}
```

In fact, because of this check, it seems like it should be safe to call `roaring64_iterator_advance` repeatedly even after reaching the end of the bitmap.

_Originally posted by @Dr-Emann in https://github.com/RoaringBitmap/CRoaring/pull/558#discussion_r1464188393_

SLieve:
Fixes #574.

Previously calling `roaring64_iterator_previous` after `roaring64_iterator_advance` returned false would not move the iterator. With this commit, this behavior and its inverse are both possible.

I chose to implement this by adding a new field in the iterator to indicate saturation. Another option would be to allow the ART iterator to allow moving from beyond the start / end, but that would require changing the index to a signed integer and thus take up more space.

### DIFF
--- a/src/roaring64.c
+++ b/src/roaring64.c
@@ -48,6 +48,12 @@ typedef struct roaring64_iterator_s {
 
     uint64_t value;
     bool has_value;
+
+    // If has_value is false, then the iterator is saturated. This field
+    // indicates the direction of saturation. If true, there are no more values
+    // in the forward direction. If false, there are no more values in the
+    // backward direction.
+    bool saturated_forward;
 } roaring64_iterator_t;
 
 // Splits the given uint64 key into high 48 bit and low 16 bit components.
@@ -126,6 +132,8 @@ static inline roaring64_iterator_t *roaring64_iterator_init_at(
         } else {
             roaring64_iterator_init_at_leaf_last(it);
         }
+    } else {
+        it->saturated_forward = first;
     }
     return it;
 }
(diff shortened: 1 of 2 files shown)


Dr-Emann (approved):

Dr-Emann:
@SLieve I believe with this implementation, it _is_ safe to call `roaring64_iterator_advance` after the iterator has already reached the end (and vice-versa) unlike the [documentation says](https://github.com/RoaringBitmap/CRoaring/blob/cf56e667f86ece017a42bc6db70643625ea49270/include/roaring/roaring64.h#L601-L602)

Is this something we want to document, or is it better to leave it open to allow us to make it unsafe in the future? (Personally, in the rust bindings, I'm currently keeping track of "which end did we last go over" as well, and it would be nice to avoid having to keep that extra state if we're willing to guarantee it's safe to do so.)

SLieve:
For consistency across the API I would lean toward disallowing it (as we disallow it in the 32-bit version), but I don't feel strongly. Alternatively we can also make it safe in the 32-bit version, which sounds nice.

PR #585, opened 2024-02-03
r64 iterator: allow moving backward with move_equalorlarger

### LINKED ISSUE #575: Roaring64 iterator's move_equalorlarger can only go forward
`roaring_uint32_iterator_move_equalorlarger` is able to move the iterator forward/backward. For consistency, we should make both be able to go in both directions

_Originally posted by @Dr-Emann in https://github.com/RoaringBitmap/CRoaring/issues/558#issuecomment-1907301009_

SLieve:
Fixes #575. This adds the ability to move backward to both the ART iterator and the roaring64 iterator.

### DIFF
--- a/include/roaring/art/art.h
+++ b/include/roaring/art/art.h
@@ -149,8 +149,6 @@ bool art_iterator_prev(art_iterator_t *iterator);
 
 /**
  * Moves the iterator forward to a key equal to or greater than the given key.
- * Assumes the given key is greater or equal to the current position of the
- * iterator.
  */
 bool art_iterator_lower_bound(art_iterator_t *iterator,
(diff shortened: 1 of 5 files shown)


Dr-Emann (approved):

Dr-Emann:
@SLieve Should `roaring64_iterator_move_equalorlarger` work if the iterator is exhausted?

E.g. should this work (as long as the bitmap is not empty)?:

```c
while (roaring64_iterator_advance(it))
    ;
assert_true(roaring64_iterator_move_equalorlarger(it, 0));
```

SLieve:
Interesting question, I don't think see a reason why it shouldn't work so let's make it work.

Issue #589, opened 2024-02-09
Copying a Roaring64MapSetBitForwardIterator into an iterator from another map does not work


tudor:
```
Roaring64Map m1;
auto it = m1.begin();
// ...
Roaring64Map m2;
it = m2.begin();
```

This leads to trouble, because `Roaring64MapSetBitForwardIterator`'s copy assignment does not copy the map reference. So if you then use `it.move(<value>)`, `map_iter` is going to point into the wrong map (`m1`).

tudor:
Fix in https://github.com/RoaringBitmap/CRoaring/pull/590

PR #590, opened 2024-02-09
Change map when copying Roaring64MapSetBitForwardIterator

tudor:
See https://github.com/RoaringBitmap/CRoaring/issues/589.

Fixed by turning the map reference into a pointer, in which case the default copy assignment operator will do the right thing. (`std::reference_wrapper` would also have worked, but I find that way too pedantic.)

Now that we did this, we don't need `map_begin` and `map_end` any more -- just call `p->cbegin()` and `p->cend()` as needed. (map iterators are very cheap to construct)

Now that we did this, we don't need the difference between `Roaring64MapSetBitForwardIterator` and `Roaring64MapSetBitBiDirectionalIterator` -- I made`Roaring64MapSetBitBiDirectionalIterator` the only iterator and added typedefs as needed for backwards compatibility. 

Added a test, all tests pass.

### DIFF
--- a/cpp/roaring64map.hh
+++ b/cpp/roaring64map.hh
@@ -31,8 +31,9 @@ namespace roaring {
 
 using roaring::Roaring;
 
-class Roaring64MapSetBitForwardIterator;
 class Roaring64MapSetBitBiDirectionalIterator;
+typedef Roaring64MapSetBitBiDirectionalIterator
+    Roaring64MapSetBitForwardIterator;
 
 class Roaring64Map {
     typedef api::roaring_bitmap_t roaring_bitmap_t;
@@ -1553,9 +1554,8 @@ class Roaring64Map {
         return result;
     }
 
-    friend class Roaring64MapSetBitForwardIterator;
     friend class Roaring64MapSetBitBiDirectionalIterator;
-    typedef Roaring64MapSetBitForwardIterator const_iterator;
+    typedef Roaring64MapSetBitBiDirectionalIterator const_iterator;
     typedef Roaring64MapSetBitBiDirectionalIterator
         const_bidirectional_iterator;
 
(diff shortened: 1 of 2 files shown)


tests/cpp_unit.cpp
```diff
@@ -2146,6 +2146,18 @@ DEFINE_TEST(test_cpp_contains_range_interleaved_containers) {
     roaring.containsRange(0x1FFFF, 0x2FFFF + 2);
 }
 
+DEFINE_TEST(test_cpp_copy_map_iterator_to_different_map) {
```
SLieve, line 2149: Could you add a brief comment about the purpose of the test?

cpp/roaring64map.hh
```diff
@@ -1727,179 +1727,144 @@ class Roaring64MapSetBitForwardIterator {
...
+        return **this == *o;
+    }
+
+    bool operator!=(const Roaring64MapSetBitBiDirectionalIterator &o) const {
+        if (map_iter == p->cend() && o.map_iter == o.p->cend()) return false;
+        if (o.map_iter == o.p->cend()) return true;
+        return **this != *o;
+    }
+
+    Roaring64MapSetBitBiDirectionalIterator(const Roaring64Map &parent,
```
SLieve, line 1836: Please move to the top of the class.

cpp/roaring64map.hh
```diff
@@ -1710,14 +1710,14 @@ class Roaring64Map {
 /**
  * Used to go through the set bits. Not optimally fast, but convenient.
  */
-class Roaring64MapSetBitForwardIterator {
+class Roaring64MapSetBitBiDirectionalIterator final {
```
SLieve, line 1713: Why `final`?

cpp/roaring64map.hh
```diff
@@ -31,8 +31,9 @@ namespace roaring {
 
 using roaring::Roaring;
 
-class Roaring64MapSetBitForwardIterator;
 class Roaring64MapSetBitBiDirectionalIterator;
+typedef Roaring64MapSetBitBiDirectionalIterator
```
SLieve, line 35: I think it would be useful to leave a comment about the purpose of this typedef, given that it's no longer used elsewhere in this file.

tests/cpp_unit.cpp
tudor: Done

cpp/roaring64map.hh
tudor: Done

cpp/roaring64map.hh
tudor: Removed

cpp/roaring64map.hh
tudor: Done

tudor:
@SLieve PTAL, thank you!

SLieve (approved):

tudor:
@SLieve @lemire I can't merge this myself as I don't have write access but it's been approved. Thank you!

PR #592, opened 2024-02-12
Simplify, and optimize roaring64_bitmap_remove_range_closed

Dr-Emann:
Only visit existing nodes, rather than every possible container in the range

This makes it feasible to use `roaring64_bitmap_remove_range_closed(r, 0, UINT64_MAX)`. Before, this would have taken forever, but now, it takes proportional time to the number of containers to remove.

See #549 (the first checkbox)


SLieve (approved):
Simpler and faster, nice!

PR #596, opened 2024-02-18
r64: allow _move_equalorlarger to move from beyond the ends of the bitmap

### LINKED ISSUE #594: roaring64_iterator_move_equalorlarger cannot come back from the end
If an iterator is exhausted, it can be moved back with `roaring64_iterator_previous`, but not `roaring64_iterator_move_equalorlarger`. That seems inconsistent

```c
roaring64_bitmap_t* r1 = roaring64_bitmap_from(0);
roaring64_iterator_t *it = roaring64_iterator_create(r1);
assert_int_equal(roaring64_iterator_value(it), 0);
assert_false(roaring64_iterator_advance(it));
assert_true(roaring64_iterator_move_equalorlarger(it, 0)); // This line returns false
assert_int_equal(roaring64_iterator_value(it), 0);
assert_false(roaring64_iterator_advance(it));

roaring64_iterator_free(it);
roaring64_bitmap_free(r1);
```

_Originally posted by @Dr-Emann in https://github.com/RoaringBitmap/CRoaring/issues/585#issuecomment-1926276459_

SLieve:
Fixes #594. Previously `roaring64_iterator_move_equalorlarger` would only return a valid result if the iterator was positioned within the bitmap. With this, the iterator is allowed to start outside the bitmap.

### DIFF
--- a/src/art/art.c
+++ b/src/art/art.c
@@ -1681,6 +1681,14 @@ bool art_iterator_prev(art_iterator_t *iterator) {
 
 bool art_iterator_lower_bound(art_iterator_t *iterator,
                               const art_key_chunk_t *key) {
+    if (iterator->value == NULL) {
+        // We're beyond the end / start of the ART so the iterator does not have
+        // a valid key. Start from the root.
+        iterator->frame = 0;
+        iterator->depth = 0;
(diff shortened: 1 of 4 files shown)


Dr-Emann (approved):
Lgtm

Issue #599, opened 2024-02-29
Compiling croaring as C++ fails on Windows


SalvatorePreviti:
Compiling CRoaring on Windows (tried on Windows 2019 in CI for roaring-node) causes a compilation error due to the use of (cast){initializer} used in C roaring source code.

(roaring_container_iterator_t){ ... } does not compile in Visual Studio when compiling in C++,
it fails with C4576: a parenthesized type followed by an initializer list is a non-standard explicit type conversion syntax

For reference, https://github.com/SalvatorePreviti/roaring-node/actions/runs/8102601181/job/22145475666

I think the solution is to replace (roaring_container_iterator_t){ ... } with roaring_container_iterator_t{ ... } when compiling in C++

PR #600, opened 2024-02-29
fix C4576 compilation error when compiling CRoaring in C++ on Windows

SalvatorePreviti:
Compiling CRoaring on Windows (tried on Windows 2019 in CI for roaring-node) causes a compilation error due to the use of (cast){initializer} used in C roaring source code.

(roaring_container_iterator_t){ ... } does not compile in Visual Studio when compiling in C++,
it fails with C4576: a parenthesized type followed by an initializer list is a non-standard explicit type conversion syntax

For reference, https://github.com/SalvatorePreviti/roaring-node/actions/runs/8102601181/job/22145475666

I think the solution is to replace (roaring_container_iterator_t){ ... } with roaring_container_iterator_t{ ... } depending if compiling for C or C++.

This PR defines a macro in containers.c that is (roaring_container_iterator_t) when compiling in C and roaring_container_iterator_t when compiling in C++

issue https://github.com/RoaringBitmap/CRoaring/issues/599

### DIFF
--- a/src/containers/containers.c
+++ b/src/containers/containers.c
@@ -5,15 +5,14 @@
 #ifdef __cplusplus
 extern "C" {
 // In Windows MSVC C++ compiler, (type){init} does not compile,
-// it causes C4576: a parenthesized type followed by an initializer list is a non-standard explicit type conversion syntax
-// The correct syntax is type{init}
+// it causes C4576: a parenthesized type followed by an initializer list is a
+// non-standard explicit type conversion syntax The correct syntax is type{init}
 #define ROARING_INIT_ROARING_CONTAINER_ITERATOR_T roaring_container_iterator_t
 namespace roaring {
 namespace internal {
 #else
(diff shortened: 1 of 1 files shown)


lemire:
@SalvatorePreviti We are getting errors with the Windows runners. I'll relaunch them in case it is a fluke.

lemire:
Nope. Looks like the tests are failing under Windows.

SalvatorePreviti:
Ok I will look into it thank you

lemire:
The problem may not be with this PR. I am investigating.

lemire:
The problem wasn't there 2 weeks ago, but it may not be caused by your PR:

Screenshot 2024-03-01 at 10 08 24 AM

lemire:
Oh. Right. Now I remember.

lemire:
It is an issue identified by @Dr-Emann

We think it is a visual studio bug.

lemire:
I am going to merge your PR because I do not think that the failure is related.

lemire:
See https://github.com/RoaringBitmap/CRoaring/pull/603

PR #641, opened 2024-07-04
Only ignore amalgamation files at the root

### LINKED ISSUE #637: gitignore tries to ignore src/roaring.c, etc
> Oh, the amalgamated file is $ROOT/roaring.c instead of src/roaring.c. The .gitignore file functions correctly in Git, but tools like fd and ripgrep somehow also ignore src/roaring.c. It will work if we explicitly specify them like this in .gitignore:

```
/amalgamation_demo.c
/amalgamation_demo.cpp
/roaring.c
/roaring.h
/roaring.hh
```

_Originally posted by @amosbird in https://github.com/RoaringBitmap/CRoaring/pull/179#discussion_r1654803111_

Dr-Emann:
Currently, the gitignore tries to ignore src/roaring.c

This actually only works in git because git only considers the gitignore when it would _add_ a file to be tracked, changes to existing files already in git are tracked regardless of what's in .gitignore.

Fixes #637

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -321,11 +321,12 @@ paket-files/
 __pycache__/
 *.pyc
 
-amalgamation_demo
-amalgamation_demo.c
-amalgamation_demo.cpp
-roaring.c
-roaring.h
-roaring.hh
+# Exclude amalgamation created files
+/amalgamation_demo
+/amalgamation_demo.c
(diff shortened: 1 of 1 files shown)


lemire (approved):
LGTM

PR #720, opened 2025-07-25
Fix C++ include paths for CMake FetchContent

brawer:
Fixes https://github.com/RoaringBitmap/CRoaring/issues/719.

Before this change, the C++ headers were available as `#include <roaring/roaring.hh>`
if CRoaring was installed as a system package. However, if a user project pulled in
CRoaring as a dependency via CMake FetchContent, it had to include C++ headers
as `#include <roaring.hh>`. This was different to the C headers, which were consistently
available as `#include <roaring/roaring.h>` no matter whether CRoaring was installed
as a system package or brought into a user project by CMake.

After this change, user cocde can consistently include the C++ headers from the same place.

To test the changes of this pull request, put the following `CMakeLists.txt` in a new, empty directory:

```cmake
cmake_minimum_required(VERSION 3.16)
project(foo)
include(FetchContent)

FetchContent_Declare(
    roaring
    GIT_REPOSITORY https://github.com/brawer/CRoaring.git
    GIT_TAG fix-cpp-include
    FIND_PACKAGE_ARGS)

FetchContent_MakeAvailable(roaring)

add_executable(foo main.cc)
target_link_libraries(foo PRIVATE roaring)
```

And the following in `main.cc`:

```c++
#include <iostream>
#include "roaring/roaring.hh"

int main() {
  roaring::Roaring r1;
  std::cout << "cardinality = " << r1.cardinality() << std::endl;
  return 0;
}
```

Then, make sure that CRoaring is **not** installed on the system, and build the project. For example, on macOS with the Homebrew package manager, do this:

```sh
$ brew uninstall croaring
$ cmake -B build-dir && cmake --build build-dir && ./build-dir/foo
[...]
cardinality = 0
```

Before the changes in this pull request, the build would have failed. See https://github.com/RoaringBitmap/CRoaring/issues/719 for additional details.


lemire:
It looks like you implemented what I had in mind.

I expect that this is a good PR. Running tests.

brawer:
The [fuzzer build error](https://github.com/RoaringBitmap/CRoaring/actions/runs/16514257899/job/46705925328?pr=720#step:4:324) is about a `#include "cpp/roaring.hh"` on line 19 of `src/croaring_fuzzer_cc.cc`. However, in my pull request, I did change line 19 of that file to `#include "cpp/roaring/roaring.hh"`. Why does the CI build try to compile the old version, before the change?

lemire:
I expect that the PR is good. I will issue a release next week.

PR #763, opened 2025-11-18
move fuzz/build.sh to oss-fuzz to fix fuzzing build error

xingyaner:
The fix involves moving the fuzz/build.sh file to the oss-fuzz/projects/croaring/ directory and applying specific modifications to correct the fuzzing build failures.  I will subsequently submit a Pull Request to the oss-fuzz repository to implement this solution.
[https://github.com/RoaringBitmap/CRoaring/issues/762](url)
[https://oss-fuzz-build-logs.storage.googleapis.com/log-73d6d8bb-72ed-4d72-b403-5b1ef5afb33b.txt](url)


xingyaner:
#758 
Of course, I think it's necessary to explain clearly.  The root cause of the problem is as follows:

In Commit 8711167, a new line of code "set(CMAKE_INTERPROCEDURAL_OPTIMIZATION TRUE)" was added to CRoaring 's ‎CMakeLists.txt.This causes the format of the libroaring.a file generated by CMake calling the Clang compiler to change from ELF to LLVM Bitcode.

  1. In the October 28th build, since libroaring.a is in ELF format, the OSS-Fuzz linker /usr/bin/ld can read and process it, so the fuzzing build is displayed as successful (in fact, there is still a hidden danger, please see the detailed information below).
  2. In the October 29th build, the linker failed to understand the Bitcode format, resulting in a failed build.

If we delete this line of code, it will terminate the IPO/LTO and restore the fuzzing build to the "successful" state.  But more importantly, this state is imperfect.

**Detailed information**
The following are the logs of successful builds on 10.28 and failed builds on 10.29, as well as the logs and related information of fuzzing builds performed locally using the environment of 10.28 and the new solution respectively.  With this information, you can easily reproduce them.

  1. As shown in the green part of "fuzzing build log of 10.10 commit", although the link was successful, CRoaring/fuzz/build.sh caused CMake to use the default Release configuration (without instrumentation) to compile libroaring.a.  This means that the library code does not have Sanitizer protection at runtime(Partial Instrumentation), and Fuzzing tests cannot detect potential memory errors that may occur inside the library.
  2. As shown in the red part of "the fuzzing build log of new scheme", the clang command explicitly contains a series of instrumentation flags such as -fsanitize=address, ensuring that the Fuzzer has complete detection capabilities.

[10.28 success log-37eb6a44-8a5e-47ff-85cc-646ddbd38bd9.txt](https://github.com/user-attachments/files/23647208/10.28.success.log-37eb6a44-8a5e-47ff-85cc-646ddbd38bd9.txt)
[10.29 error log-189de54d-cb21-45e9-8c8f-37c767da42a3.txt](https://github.com/user-attachments/files/23647215/10.29.error.log-189de54d-cb21-45e9-8c8f-37c767da42a3.txt)
[fuzzing build log of 10.10 commit.docx](https://github.com/user-attachments/files/23647276/fuzzing.build.log.of.10.10.commit.docx)
[the fuzzing build log of new scheme.docx](https://github.com/user-attachments/files/23647281/the.fuzzing.build.log.of.new.scheme.docx)

xingyaner:
#758 
Of course, I think it's necessary to explain clearly. The root cause of the problem is as follows:
In Commit 8711167, a new line of code "set(CMAKE_INTERPROCEDURAL_OPTIMIZATION TRUE)" was added to CRoaring 's ‎CMakeLists.txt.This causes the format of the libroaring.a file generated by CMake calling the Clang compiler to change from ELF to LLVM Bitcode.

  1. In the October 28th build, since libroaring.a is in ELF format, the OSS-Fuzz linker /usr/bin/ld can read and process it, so the fuzzing build is displayed as successful .
  2.In the October 29th build, the linker failed to understand the Bitcode format, resulting in a failed build.

Error log shows "/usr/bin/ld: /src/croaring/build-dir/src/libroaring.a: error adding symbols: file format not recognized
clang++: error: linker command failed with exit code 1"

If we delete this line of code, it will terminate the IPO/LTO and restore the fuzzing build to the "successful" state. This is the simplest method.

PR #792, opened 2026-03-22
Enable strict compiler warnings with -Werror

maxwbuckley:
## Summary

- Adds a `ROARING_WERROR` CMake option (OFF by default) that enables `-Werror -Wundef -Wnull-dereference -Wmissing-prototypes` for developer builds
- Default build flags are unchanged (`-Wall` only), so downstream users are unaffected
- Fixes all warnings triggered by the stricter flags

## Changes

**Build system:**
- `CMakeLists.txt`: Added `ROARING_WERROR` option (OFF by default)
- `FindOptions.cmake`: `-Werror`, `-Wundef`, `-Wnull-dereference`, `-Wmissing-prototypes`, and amalgamation warning suppressions are all gated behind `ROARING_WERROR`
- `FindOptions.cmake`: Centralized warning flag variables (`C_EXTRA_WARNING_FLAGS`, `AMALG_EXTRA_WARNING_FLAGS`) instead of scattering flags across individual CMake files

**Warning fixes:**
- Made `container_unwrap_shared`/`container_mutable_unwrap_shared` non-static inline to fix static-in-non-static-inline warning
- Added `static` to file-local functions (`art_node_printf`, `align_size`, `_avx2/_scalar_run_container_to_uint32_array`)
- Added NULL checks after allocation calls in `bitset_container_offset`, `run_container_offset`, and `roaring64_flip_leaf`
- Initialized `leaf` variable to fix maybe-uninitialized warning
- Guarded preprocessor macros with `defined()` for `-Wundef`: `__cplusplus`, `__STDC_VERSION__`, `_MSC_VER`, `__LITTLE_ENDIAN__`, `__BIG_ENDIAN__`, `__CHAR_BIT__`, `ROARING_UNSAFE_FROZEN_TESTS`

## Test plan

- [x] Full clean build with `-DROARING_WERROR=ON` passes (library + tests + benchmarks + amalgamation)
- [x] Full clean build with default flags (no `ROARING_WERROR`) passes
- [x] All 66 unit tests pass on both build configurations
- [ ] CI passes on Linux (GCC), macOS (Clang), and Windows (MSVC)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -45,6 +45,7 @@ option(ROARING_LINK_STATIC "Link executables (tests, benchmarks) statically" OFF
 option(ROARING_BUILD_LTO "Build library with Link Time Optimization" OFF)
 option(ROARING_BUILD_C_AS_CPP "Build library C files using C++ compilation" OFF)
 option(ROARING_BUILD_C_TESTS_AS_CPP "Build test C files using C++ compilation" OFF)
+option(ROARING_WERROR "Treat warnings as errors (developer only)" OFF)
 option(ROARING_SANITIZE "Sanitize addresses" OFF)
 option(ROARING_SANITIZE_THREADS "Sanitize threads" OFF)
 option(ROARING_SANITIZE_UNDEFINED "Sanitize undefined behaviors" OFF)
@@ -146,21 +147,16 @@ if(ENABLE_ROARING_TESTS AND BASH AND NOT EMSCRIPTEN)
 
   add_executable(amalgamate_demo $<BUILD_INTERFACE:${CMAKE_CURRENT_BINARY_DIR}/amalgamation_demo.c>)
   target_link_libraries(amalgamate_demo croaring-singleheader-include-source)
-  # The amalgamation compiles everything in a single translation unit, which
-  # gives GCC deeper interprocedural visibility and triggers false-positive
-  # null-dereference warnings through allocation-failure paths. Suppress
-  # this warning for amalgamation targets only; the library itself is still
-  # checked with -Wnull-dereference in separate compilation units.
-  if(NOT MSVC)
-    target_compile_options(amalgamate_demo PRIVATE -Wno-null-dereference)
+  if(AMALG_EXTRA_WARNING_FLAGS)
+    target_compile_options(amalgamate_demo PRIVATE ${AMALG_EXTRA_WARNING_FLAGS})
   endif()
   add_test(amalgamate_demo amalgamate_demo)
 
   add_library(croaring-singleheader-source-lib $<BUILD_INTERFACE:${CMAKE_CURRENT_BINARY_DIR}/roaring.c>)
   set_target_properties(croaring-singleheader-source-lib PROPERTIES WINDOWS_EXPORT_ALL_SYMBOLS YES)
   target_include_directories(croaring-singleheader-source-lib PUBLIC $<BUILD_INTERFACE:${CMAKE_CURRENT_BINARY_DIR}>)
-  if(NOT MSVC)
-    target_compile_options(croaring-singleheader-source-lib PRIVATE -Wno-null-dereference)
+  if(AMALG_EXTRA_WARNING_FLAGS)
+    target_compile_options(croaring-singleheader-source-lib PRIVATE ${AMALG_EXTRA_WARNING_FLAGS})
   endif()
 
   add_executable(amalgamate_demo_cpp $<BUILD_INTERFACE:${CMAKE_CURRENT_BINARY_DIR}/amalgamation_demo.cpp>)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -146,11 +146,22 @@ if(ENABLE_ROARING_TESTS AND BASH AND NOT EMSCRIPTEN)
 
   add_executable(amalgamate_demo $<BUILD_INTERFACE:${CMAKE_CURRENT_BINARY_DIR}/amalgamation_demo.c>)
   target_link_libraries(amalgamate_demo croaring-singleheader-include-source)
+  # The amalgamation compiles everything in a single translation unit, which
+  # gives GCC deeper interprocedural visibility and triggers false-positive
+  # null-dereference warnings through allocation-failure paths. Suppress
+  # this warning for amalgamation targets only; the library itself is still
+  # checked with -Wnull-dereference in separate compilation units.
+  if(NOT MSVC)
+    target_compile_options(amalgamate_demo PRIVATE -Wno-null-dereference)
+  endif()
   add_test(amalgamate_demo amalgamate_demo)
 
   add_library(croaring-singleheader-source-lib $<BUILD_INTERFACE:${CMAKE_CURRENT_BINARY_DIR}/roaring.c>)
   set_target_properties(croaring-singleheader-source-lib PROPERTIES WINDOWS_EXPORT_ALL_SYMBOLS YES)
   target_include_directories(croaring-singleheader-source-lib PUBLIC $<BUILD_INTERFACE:${CMAKE_CURRENT_BINARY_DIR}>)
+  if(NOT MSVC)
+    target_compile_options(croaring-singleheader-source-lib PRIVATE -Wno-null-dereference)
+  endif()
 
   add_executable(amalgamate_demo_cpp $<BUILD_INTERFACE:${CMAKE_CURRENT_BINARY_DIR}/amalgamation_demo.cpp>)
   target_link_libraries(amalgamate_demo_cpp croaring-singleheader-include-source croaring-singleheader-source-lib)
--- a/include/roaring/containers/containers.h
+++ b/include/roaring/containers/containers.h
@@ -90,7 +90,7 @@ container_t *shared_container_extract_copy(shared_container_t *container,
                                            uint8_t *typecode);
 
 /* access to container underneath */
-static inline const container_t *container_unwrap_shared(
+inline const container_t *container_unwrap_shared(
     const container_t *candidate_shared_container, uint8_t *type) {
     if (*type == SHARED_CONTAINER_TYPE) {
         *type = const_CAST_shared(candidate_shared_container)->typecode;
@@ -102,8 +102,8 @@ static inline const container_t *container_unwrap_shared(
 }
 
 /* access to container underneath */
-static inline container_t *container_mutable_unwrap_shared(container_t *c,
-                                                           uint8_t *type) {
+inline container_t *container_mutable_unwrap_shared(container_t *c,
+                                                    uint8_t *type) {
     if (*type == SHARED_CONTAINER_TYPE) {  // the passed in container is shared
         *type = CAST_shared(c)->typecode;
         assert(*type != SHARED_CONTAINER_TYPE);
--- a/include/roaring/misc/configreport.h
+++ b/include/roaring/misc/configreport.h
@@ -186,14 +186,14 @@ static inline void tellmeall() {
                (long unsigned int)sizeof(size_t),
                (long unsigned int)sizeof(int));
     }
-#if defined(__LITTLE_ENDIAN__) && __LITTLE_ENDIAN__
+#if defined(__LITTLE_ENDIAN__)
 // This is what we expect!
 // printf("you have little endian machine");
 #endif
-#if defined(__BIG_ENDIAN__) && __BIG_ENDIAN__
+#if defined(__BIG_ENDIAN__)
     printf("you have a big endian machine");
 #endif
-#if defined(__CHAR_BIT__) && __CHAR_BIT__
+#if defined(__CHAR_BIT__)
     if (__CHAR_BIT__ != 8) printf("on your machine, chars don't have 8bits???");
 #endif
     if (computecacheline() != 64)
@@ -217,14 +217,14 @@ static inline void tellmeall() {
                (long unsigned int)sizeof(size_t),
                (long unsigned int)sizeof(int));
     }
-#if defined(__LITTLE_ENDIAN__) && __LITTLE_ENDIAN__
+#if defined(__LITTLE_ENDIAN__)
 // This is what we expect!
 // printf("you have little endian machine");
 #endif
-#if defined(__BIG_ENDIAN__) && __BIG_ENDIAN__
+#if defined(__BIG_ENDIAN__)
     printf("you have a big endian machine");
 #endif
-#if defined(__CHAR_BIT__) && __CHAR_BIT__
+#if defined(__CHAR_BIT__)
     if (__CHAR_BIT__ != 8) printf("on your machine, chars don't have 8bits???");
 #endif
 }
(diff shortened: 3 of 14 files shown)


tools/cmake/FindOptions.cmake
```diff
@@ -24,7 +24,7 @@ endif()
 endif()
 
 if(NOT MSVC)
-set(WARNING_FLAGS "-Wall")
+set(WARNING_FLAGS "-Wall -Werror -Wnull-dereference -Wundef")
```
madscientist: No.

It's wrong to enable -Werror by default.  Just because the code compiles FOR YOU with YOUR COMPILER version without warnings, does not mean it will compile for other people with their compiler version without warnings.  Plus, new compiler versions always come along that add new warnings.

A package like CRoaring should never enable -Werror as a default setting because that prevents the package from compiling on other peoples' systems unless they go through the effort of fixing all the warnings generated on their system, first.

The -Werror option should be a developer-only (by developer here I mean, CRoaring developer) option that is disabled by default, and can be enabled as a CMake flag for those who want it.

CMakeLists.txt
```diff
@@ -146,11 +146,22 @@ if(ENABLE_ROARING_TESTS AND BASH AND NOT EMSCRIPTEN)
 
   add_executable(amalgamate_demo $<BUILD_INTERFACE:${CMAKE_CURRENT_BINARY_DIR}/amalgamation_demo.c>)
   target_link_libraries(amalgamate_demo croaring-singleheader-include-source)
+  # The amalgamation compiles everything in a single translation unit, which
+  # gives GCC deeper interprocedural visibility and triggers false-positive
+  # null-dereference warnings through allocation-failure paths. Suppress
+  # this warning for amalgamation targets only; the library itself is still
+  # checked with -Wnull-dereference in separate compilation units.
+  if(NOT MSVC)
+    target_compile_options(amalgamate_demo PRIVATE -Wno-null-dereference)
```
madscientist: This is bad organization.

Instead of this, a variable should be defined in the same place as the WARNING_FLAGS options are defined, containing extra warning options added to amalgamation compiles for a given compiler.  Then that variable should be applied to the amalgamation targets.

include/roaring/misc/configreport.h
```diff
@@ -186,14 +186,14 @@ static inline void tellmeall() {
                (long unsigned int)sizeof(size_t),
                (long unsigned int)sizeof(int));
     }
-#if __LITTLE_ENDIAN__
+#if defined(__LITTLE_ENDIAN__) && __LITTLE_ENDIAN__
```
madscientist: I don't believe these are needed.  The __LITTLE_ENDIAN__  __BIG_ENDIAN__ macros are always defined to be a non-0 value, which can be compared to __BYTE_ORDER__.

include/roaring/misc/configreport.h
```diff
@@ -217,14 +217,14 @@ static inline void tellmeall() {
...
 // This is what we expect!
 // printf("you have little endian machine");
 #endif
-#if __BIG_ENDIAN__
+#if defined(__BIG_ENDIAN__) && __BIG_ENDIAN__
     printf("you have a big endian machine");
 #endif
-#if __CHAR_BIT__
+#if defined(__CHAR_BIT__) && __CHAR_BIT__
```
madscientist: This is REALLY not needed since the definition of this value is the number of bits in a char which obviously cannot be 0.

Personally I would remove the second part and keep just the #if defined(__CHAR_BIT__) but you definitely don't need both

include/roaring/portability.h
```diff
@@ -584,7 +584,7 @@ static inline uint32_t croaring_refcount_get(const croaring_refcount_t *val) {
 
 // We want to initialize structs to zero portably (C and C++), without
 // warnings. We can do mystruct s = CROARING_ZERO_INITIALIZER;
-#if __cplusplus
+#if defined(__cplusplus) && __cplusplus
```
madscientist: Similar to __CHAR_BIT__ this is a standard macro and can't ever be zero.  I also think it makes more sense to keep the defined() version, but definitely you don't need both so just choose one.

src/containers/containers.c
```diff
@@ -52,8 +52,7 @@ extern bool container_iterator_prev(const container_t *c, uint8_t typecode,
                                     uint16_t *value);
 extern bool container_contains(
     const container_t *c, uint16_t val,
-    uint8_t typecode  // !!! should be second argument?
-);
+    uint8_t typecode);
```
madscientist: Don't understand this change

src/containers/containers.c
```diff
@@ -163,7 +162,7 @@ extern inline container_t *container_add(container_t *c, uint16_t val,
                                          uint8_t *new_typecode);
 
 extern inline bool container_contains(const container_t *c, uint16_t val,
-                                      uint8_t typecode);  // !!! 2nd arg?
+                                      uint8_t typecode);
```
madscientist, line 167: why?

src/isadetection.c
```diff
@@ -54,7 +54,7 @@ POSSIBILITY OF SUCH DAMAGE.
 // on these systems. It seems that ClangCL is not affected.
 // https://github.com/RoaringBitmap/CRoaring/pull/603
 #ifndef __clang__
-#if _MSC_VER == 1938
+#if defined(_MSC_VER) && _MSC_VER == 1938
```
madscientist, line 57: Definitely not needed.  If __MSC_VER is not defined it will be substituted with 0 which won't match.

src/CMakeLists.txt
```diff
@@ -63,6 +63,9 @@ target_link_libraries(roaring PUBLIC "$<BUILD_INTERFACE:roaring-headers>")
 target_link_libraries(roaring PUBLIC "$<BUILD_INTERFACE:roaring-headers-cpp>")
 
 target_include_directories(roaring PRIVATE ${CMAKE_CURRENT_SOURCE_DIR})
+if(NOT MSVC)
+  target_compile_options(roaring PRIVATE $<$<COMPILE_LANGUAGE:C>:-Wmissing-prototypes>)
```
madscientist: This seems like it belongs somewhere more global and not in an individual cmake file.

tests/cpp_unit.cpp
```diff
@@ -1476,7 +1476,7 @@ DEFINE_TEST(test_cpp_frozen_64) {
     roaring_aligned_free(buf);
 }
 
-#if ROARING_UNSAFE_FROZEN_TESTS
+#if defined(ROARING_UNSAFE_FROZEN_TESTS) && ROARING_UNSAFE_FROZEN_TESTS
```
madscientist: I don't see how any of these are needed: the cmake file ensures that if the value is defined it's defined to be 1.

include/roaring/containers/containers.h
```diff
@@ -90,7 +90,7 @@ container_t *shared_container_extract_copy(shared_container_t *container,
                                            uint8_t *typecode);
 
 /* access to container underneath */
-static inline const container_t *container_unwrap_shared(
+inline const container_t *container_unwrap_shared(
```
lemire, line 93: Why would we need external linkage for this function?

include/roaring/containers/containers.h
```diff
@@ -102,8 +102,8 @@ static inline const container_t *container_unwrap_shared(
 }
 
 /* access to container underneath */
-static inline container_t *container_mutable_unwrap_shared(container_t *c,
-                                                           uint8_t *type) {
+inline container_t *container_mutable_unwrap_shared(container_t *c,
```
lemire, line 105: Why would we need external linkage for this function? (same question)

src/containers/bitset.c
```diff
@@ -176,6 +176,9 @@ void bitset_container_offset(const bitset_container_t *c, container_t **loc,
 
     if (loc != NULL) {
         bc = bitset_container_create();
+        if (bc == NULL) {
```
lemire, line 179: Can bc be null here?

tools/cmake/FindOptions.cmake
maxwbuckley: Thank you — you're absolutely right, and I appreciate the thorough explanation. I've moved `-Werror` (along with `-Wundef`, `-Wnull-dereference`, and `-Wmissing-prototypes`) behind a new `ROARING_WERROR` CMake option that defaults to OFF. The default build flags remain unchanged (`-Wall` only), so downstream users and package builders are unaffected. Developers can opt in with `-DROARING_WERROR=ON`.

CMakeLists.txt
maxwbuckley: Good point, thanks. I've centralized the warning flag variables in `FindOptions.cmake` — `C_EXTRA_WARNING_FLAGS` for `-Wmissing-prototypes` and `AMALG_EXTRA_WARNING_FLAGS` for the amalgamation suppressions. The amalgamation targets now reference these variables instead of hardcoding flags inline.

include/roaring/misc/configreport.h
maxwbuckley: You're right — these macros are always non-zero when defined. Simplified to just `#if defined(X)`, dropping the redundant `&& X`. The `defined()` guard itself is still needed for `-Wundef` when building with `-DROARING_WERROR=ON`, since these macros aren't defined on all platforms.

include/roaring/misc/configreport.h
maxwbuckley: Agreed — simplified to `#if defined(__CHAR_BIT__)`. The value can never be 0 as you noted.

include/roaring/portability.h
maxwbuckley: Agreed — simplified to `#if defined(__cplusplus)`. The `defined()` check is still needed because this header is compiled in both C and C++ modes, and `-Wundef` (under `ROARING_WERROR`) would warn about the undefined macro in C mode.

src/containers/containers.c
maxwbuckley: Sorry about the confusion — this was just removing a stale `// !!! should be second argument?` comment. I've reverted it since it's unrelated to the warning fixes.

src/containers/containers.c
maxwbuckley: Same — was removing a `// !!! 2nd arg?` comment. Reverted.

src/isadetection.c
maxwbuckley: You're right that behavior-wise the `defined()` check is redundant since 0 won't match 1938. However, this code is compiled on GCC/Clang too (guarded only by `#ifndef __clang__`), where `_MSC_VER` is undefined. With `-Wundef` enabled (under `ROARING_WERROR`), the preprocessor warns about using an undefined macro in an `#if` expression. The `defined()` guard silences that. Happy to hear if you'd prefer a different approach.

src/CMakeLists.txt
maxwbuckley: Agreed — moved to `FindOptions.cmake` alongside the other warning flags, exposed via the `C_EXTRA_WARNING_FLAGS` variable.

tests/cpp_unit.cpp
maxwbuckley: You're right that the `&& ROARING_UNSAFE_FROZEN_TESTS` part is redundant since CMake defines it to 1. Simplified to `#if defined(ROARING_UNSAFE_FROZEN_TESTS)`. The `defined()` check is still needed because when the option is OFF the macro isn't defined at all, and `-Wundef` (under `ROARING_WERROR`) would warn.

include/roaring/containers/containers.h
maxwbuckley: Good question. These functions are called from other non-static `inline` functions in the same header (e.g., `container_contains`). With strict warnings, GCC warns about a `static` function being used inside a non-static `inline` function — the static function may not be visible in all translation units where the inline function could be emitted. Removing `static` resolves this while maintaining the same inlining behavior.

include/roaring/containers/containers.h
maxwbuckley: Same reason as above — called from non-static `inline` functions in this header, so it needs matching linkage.

src/containers/bitset.c
maxwbuckley: Yes — `bitset_container_create()` calls `roaring_malloc()` internally and returns NULL on allocation failure. The NULL check is defensive against that. Without it, `-Wnull-dereference` (under `ROARING_WERROR`) flags the subsequent `bc->words` / `bc->cardinality` accesses.

lemire:
@maxwbuckley Thanks for the pull request.

I have pushed a commit with credit to you: 

https://github.com/RoaringBitmap/CRoaring/commit/8808dcebefe7fd00f7fa173a09243efc6e83a31a

Otherwise, we are not going to go with this PR.

We can already do:

```
cmake -B build -DCMAKE_COMPILE_WARNING_AS_ERROR=ON
```

or, on older versions of cmake:

```
cmake -B build -DCMAKE_CXX_FLAGS="-Werror"
```

and so forth.

It is very much not an objective to compile warning free under all possible static analyzers.

The PR makes several changes to the code which, I think, you are doing to satisfy some static analysis warnings. But that's not what we want to do.


What we want to do is look at the warning and then understand it. So before we make changes, we want to *know* that the change is justified.

« I wanted to silence a warning » is not a good enough reason.

It is easy enough to just as Claude or any AI to jack up the warning settings and then work at fixing the warnings, but what does it buy us?

Code changes should mostly fall into one of three categories:

1. Better performance.
2. New features.
3. Bug fixes.

If there are components of this PR that fall into these categories and I missed it, please reopen a new PR.

lemire:
@maxwbuckley The assumption in CRoaring is that allocations work. If a null is returned, then things will definitely fall apart (you'll get a crash). Whether silently continuing or crashing is the best option can be debated but we have no test at all for these conditions currently.

We toyed with the idea of catching allocation failures... but this was never resolved. If you are interested in working on this problem, help is certainly welcome... We must first decide how we handle the allocations failure.


Note that on many systems that we care about, the `malloc` will not return null typically. It will succeed, and the failure will come later when you access the memory. 

In CRoaring, you can setup your own memory allocators...

```c
roaring_memory_t my_memory = {
        .malloc         = my_malloc,
        .realloc        = my_realloc,
        .calloc         = my_calloc,
        .free           = my_free,
        .aligned_malloc = my_aligned_malloc,
        .aligned_free   = my_aligned_free
    };

    roaring_init_memory_hook(&my_memory);  
```

And then one could handle the memory failures. In many instances, it should be a fatal crash...

So you can do something of the sort...

```C
void *xmalloc(size_t size)
{
    void *ptr = malloc(size);
    if (ptr == NULL) {
        fprintf(stderr, "Fatal: out of memory (tried to allocate %zu bytes)\n", size);
        exit(EXIT_FAILURE);        // or abort() if you prefer
        /* or: _Exit(EXIT_FAILURE); */
    }
    return ptr;
}
```

Catching and ignoring NULLs out of memory allocations is not, on its own, better than a crash.

But, coming back... consider that returning NULL is not the standard...

Capture d’écran, le 2026-03-24 à 11 00 37

So it is understood that, typically, an out-of-memory should just crash the system (you can put hooks if you need cleaning).

How CRoaring shoudl behave is up to debates at this point.

PR #795, opened 2026-04-01
Added a mean allocator to deserialization tests

Dr-Emann:
Trying to debug a panic in the rust bindings CI, that seems to be failing on a misaligned simd read on windows

### DIFF
--- a/tests/robust_deserialization_unit.c
+++ b/tests/robust_deserialization_unit.c
@@ -8,13 +8,72 @@
 #include <stdlib.h>
 #include <string.h>
 
+#include <roaring/memory.h>
 #include <roaring/misc/configreport.h>
 #include <roaring/roaring.h>
 
 #include "config.h"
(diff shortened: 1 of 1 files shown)


Dr-Emann:
Hmm. It actually looks like it's a spill to the stack.

Crashing on `vmovdqa ymmword ptr [rbp-10h], ymm0`

In `avx2_harley_seal_popcount256`

Feels kinda possibly compiler bug-y. But why is it only showing up in the rust bindings test? 😒

Dr-Emann:
Shoot, I think it's https://github.com/simdjson/simdjson/issues/1963

Dr-Emann:
Maybe still worth adding something like this allocator for tests, though.

lemire:
@Dr-Emann We can merge this if you'd like.

Dr-Emann:
I think for now, I'm going to just close it

PR #798, opened 2026-04-03
add read_ranges and read_prev_ranges iterator functions

### LINKED ISSUE #776: Expose the ability to efficently get the next contigious range of set bits
**Is your feature request related to a problem? Please describe.**

Same as RoaringBitmap/roaring-rs#274.

**Describe the solution you'd like**
Basically the same as RoaringBitmap/roaring-rs#338 - A method to consume a contiguous range of values on an iterator, probably in forward or backwards direction.

It may be worth providing the ability to read a _batch_ of ranges rather than only a single range.

Maybe something like:

```c
typedef struct roaring_uint32_range_inclusive_s {
    uint32_t min;
    uint32_t max;
} roaring_uint32_range_inclusive_t;

/**
 * Reads next ${count} ranges from iterator into user-supplied ${buf}.
 * A range is defined as a maximal interval of consecutive values.
 * For example, the set {1,2,3,5,6} contains two ranges: [1..3] and [5..6].
 * Each range is represented as a struct {min,max}, both inclusive.
 *
 * Returns the number of read ranges.
 * This number can be smaller than ${count}, which means that the iterator is
 * drained.
 *
 * This function satisfies the semantics of iteration and can be used together with
 * other iterator functions.
 *  - first range will start with ${it}->current_value
 *  - after the function returns, the iterator is positioned at the next element
 *    after the end of the last returned range.
 */
uint32_t roaring_uint32_iterator_read_ranges(roaring_uint32_iterator_t *it,
                                             roaring_uint32_range_inclusive_t *buf, uint32_t count);

```

**Describe alternatives you've considered**
Currently, the best option is probably using `roaring_uint32_iterator_read`, and manually keeping track of contiguous values.

**Additional context**
Yep

**Are you willing to contribute code or documentation toward this new feature?**
It is a community-driven project.

Dr-Emann:
This adds `roaring_uint32_iterator_read_ranges` / `roaring_uint32_iterator_read_prev_ranges` to the 32-bit bitmap API, `roaring64_iterator_read_ranges` / `roaring64_iterator_read_prev_ranges` to the 64-bit bitmap API, and `read_ranges` / `read_prev_ranges` to the C++ `RoaringSetBitBiDirectionalIterator`. Each function reads up to N maximal consecutive ranges from the iterator in a single call, merging across internal container boundaries.

Closes #776

### DIFF
--- a/include/roaring/roaring.h
+++ b/include/roaring/roaring.h
@@ -1261,6 +1261,60 @@ uint32_t roaring_uint32_iterator_skip(roaring_uint32_iterator_t *it,
 uint32_t roaring_uint32_iterator_skip_backward(roaring_uint32_iterator_t *it,
                                                uint32_t count);
 
+typedef struct roaring_uint32_range_closed_s {
+    uint32_t min;
+    uint32_t max;
+} roaring_uint32_range_closed_t;
+
+/**
+ * Reads next ${count} ranges from iterator into user-supplied ${buf}.
+ * A range is defined as a maximal interval of consecutive values.
+ * For example, the set {1,2,3,5,6} contains two ranges: [1..3] and [5..6].
+ * Each range is represented as a struct {min,max}, both endpoints included.
+ * Consecutive values that span internal container boundaries are merged into
+ * a single range.
+ *
+ * Returns the number of read ranges.
+ * This number can be smaller than ${count}, which means that the iterator is
+ * drained.
+ *
+ * This function satisfies the semantics of iteration and can be used together
+ * with other iterator functions.
+ *  - first range will start with ${it}->current_value
+ *  - after the function returns, the iterator is positioned at the next element
+ *    after the end of the last returned range, or ${it}->has_value is false if
+ *    the bitmap is exhausted.
+ */
+size_t roaring_uint32_iterator_read_ranges(roaring_uint32_iterator_t *it,
+                                           roaring_uint32_range_closed_t *buf,
+                                           size_t count);
+
+/**
+ * Reads previous ${count} ranges from iterator into user-supplied ${buf}.
+ * A range is defined as a maximal interval of consecutive values.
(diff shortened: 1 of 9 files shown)


include/roaring/roaring.h
```diff
@@ -1261,6 +1261,60 @@ uint32_t roaring_uint32_iterator_skip(roaring_uint32_iterator_t *it,
 uint32_t roaring_uint32_iterator_skip_backward(roaring_uint32_iterator_t *it,
                                                uint32_t count);
 
+typedef struct roaring_uint32_range_closed_s {
+    uint32_t min;
+    uint32_t max;
+} roaring_uint32_range_closed_t;
```
lemire, line 1267: Do we want to introduce a new type? I don't mind, but it seems that it adds complexity, no?

include/roaring/roaring.h
```diff
@@ -1261,6 +1261,60 @@ uint32_t roaring_uint32_iterator_skip(roaring_uint32_iterator_t *it,
...
+ * drained.
+ *
+ * This function satisfies the semantics of iteration and can be used together
+ * with other iterator functions.
+ *  - first range will start with ${it}->current_value
+ *  - after the function returns, the iterator is positioned at the next element
+ *    after the end of the last returned range, or ${it}->has_value is false if
+ *    the bitmap is exhausted.
+ */
+size_t roaring_uint32_iterator_read_ranges(roaring_uint32_iterator_t *it,
```
lemire, line 1288: The closed range is passed by pointer? It is 8 bytes.

Dr-Emann: It's an out parameter, and is an array

lemire: Point taken.

lemire (approved):
Merge at will!

PR #799, opened 2026-04-03
Non-functional OpenZL prototype

lemire:
Long term, we want to support [OpenZL](https://openzl.org).

This is a prototype written with Claude, but Claude says that little of the SDDL format is actually implemented, so we cannot express (yet) our data format.

@Cyan4973


Cyan4973:
FYI, SDDL will receive some important upgrades for next OpenZL release, notably an ability to interpret formats with optional fields and sections, alongside substantial speed optimizations. Unfortunately, the sddl spec in this PR also requires the more advanced `scan` types, for fields which size can only be determined at run time and one by one. Notably:

```
Record RunContainer() = {
    num_runs:   UInt16LE,       # number of runs in this container
    runs:       Run[num_runs]
}
```

This capability is well within sddl scope though, and planned for later this year.

PR #802, opened 2026-04-22
builds with  Fil-C

lemire:
This PR adds support for building CRoaring with **Fil-C**, the memory-safe C compiler.

Fil-C instruments code for runtime memory safety (bounds checking, use-after-free prevention, etc.) via invisible capabilities and concurrent GC. While highly compatible, it is stricter than regular Clang/GCC in several areas, especially around:

- Runtime CPU feature detection / dynamic dispatch
- Inline assembly

### Changes

- **Use special code for CPU dispatching** when building with Fil-C.  
 
- **Disable or wrap inline assembly** that Fil-C cannot safely instrument.  
  Most of the hand-written SIMD assembly (AVX2/AVX-512) is now skipped under Fil-C.

- **Fix several warnings and minor issues** unrelated Fil-C.


include/roaring/misc/configreport.h
```diff
@@ -27,6 +27,11 @@ static inline void native_cpuid(unsigned int *eax, unsigned int *ebx,
     __asm volatile("cpuid"
                    : "=a"(*eax), "=b"(*ebx), "=c"(*ecx), "=d"(*edx)
                    : "0"(*eax), "2"(*ecx));
+#else
+    (void)eax;
+    (void)ebx;
+    (void)ecx;
+    (void)edx;
```
pizlonator, line 34: Normally in Fil-C you'd just want to use `__get_cpuid` from `<cpuid.h>` and `zxgetbv` from `<stdfil.h>`.

That's what I did to port simdutf.

But simdutf uses SIMD intrinsics. I *think* CRoaring uses inline asm instead of intrinsics? If that's right, then this change is right.

Basically:

- Fil-C supports SIMD intrinsics pretty well. I may have some missing support scatter-gather or *some* forms of masked SIMD (though even for masked SIMD, lots of those intrinsics work and I have tests for them).
- you want isadetection to take the path that uses <cpuid.h>, and change the inline asm for xgetbv to call xzgetbv from stdfil

src/roaring64.c
```diff
@@ -789,7 +789,7 @@ void roaring64_bitmap_remove_bulk(roaring64_bitmap_t *r,
         }
         if (!container_nonzero_cardinality(container2, typecode2)) {
             container_free(container2, typecode2);
-            leaf_t leaf;
+            leaf_t leaf = 0;
             bool erased = art_erase(art, high48, (art_val_t *)&leaf);
```
pizlonator, line 793: Was this necessary for Fil-C?

(Fil-C auto initializes locals to 0)

pizlonator (commented):
lgtm

src/roaring64.c
lemire: No, not necessary. This is a stupid GCC warning (false positive).

include/roaring/misc/configreport.h
lemire: > But simdutf uses SIMD intrinsics. I think CRoaring uses inline asm instead of intrinsics? If that's right, then this change is right.

We use both. We use inline assembly in some cases, but that's now disabled with Fil-C.

> Normally in Fil-C you'd just want to use __get_cpuid from <cpuid.h> and zxgetbv from <stdfil.h>.

Ah. Ok.

lemire: It works!
Updating.

lemire:
@pizlonator In our next release, we shall support building CRoaring with Fil-C.

Ideally, we would have CI tests. But... it should be trivial from now.

Issue #419, opened 2022-12-25
Fast/convenient masking operations


fbbdev:
Currently one can perform in-place _or_, _andnot_ and _xor_ operations with a range parameter through the `addRange`, `removeRange` and `flip` methods. Range-based _and_ can be simulated by two consecutive calls to `removeRange`:
```c++
    bitmap.removeRange(0, a);
    bitmap.removeRange(b + 1, bitmap.maximum() + 1);
```
It would be more convenient to have a method family like, e.g., `mask` and `maskClosed` implementing the same functionality directly.

It would also be helpful to have a family of methods for performing range-restricted updates, i.e. something like
```c++
    bitmap1.add(otherBitmap, rangeStart, rangeEnd);
    bitmap2.xor(otherBitmap, rangeStart, rangeEnd);
    bitmap3.andnot(otherBitmap, rangeStart, rangeEnd);
```
whose semantics would be
```c++
    Roaring maskedOp;
    maskedOp.addRange(rangeStart, rangeEnd);
    maskedOp &= otherBitmap;

    bitmap1 |= maskedOp;
    bitmap2 ^= maskedOp;
    bitmap3 -= maskedOp;
```
Specialized methods could probably be much faster and more memory-efficient than the mask-based solution above.

Yokiatch:
Hi, I'd like to work on this if it's still open! I was thinking of starting with the `mask` and `maskClosed` methods as a first step, since they have a clear implementation path (equivalent to two consecutive `removeRange` calls, but more convenient and potentially faster). Would that be a good starting point, or is there a preferred approach you had in mind?

lemire:
@Yokiatch We welcome pull requests.

PR #816, opened 2026-05-19
feat: add mask and maskClosed methods to Roaring

Yokiatch:
Related to #419

## Summary
Adds two convenience methods to the `Roaring` class in `roaring.hh`:

- **`mask(uint64_t min, uint64_t max)`** — keeps only values in the half-open interval `[min, max)`, equivalent to two consecutive `removeRange` calls
- **`maskClosed(uint32_t min, uint32_t max)`** — keeps only values in the closed interval `[min, max]`, equivalent to two consecutive `removeRangeClosed` calls

The method signatures follow the existing convention in the codebase: open intervals use `uint64_t` (matching `removeRange` and `addRange`), closed intervals use `uint32_t` (matching `removeRangeClosed` and `addRangeClosed`).

Both methods guard against calling `maximum()` on an empty bitmap.

## Tests
Added `test_cpp_mask` in `tests/cpp_unit.cpp` covering:
- Basic masking with values inside and outside the range
- `min == max` edge case
- Range covering all elements
- Empty bitmap
- Both `mask` and `maskClosed` variants

All 67 tests pass.

## AI Usage Disclosure
I used AI assistance during development. I reviewed and understood all code before submission and can explain my implementation choices.

### DIFF
--- a/cpp/roaring/roaring.hh
+++ b/cpp/roaring/roaring.hh
@@ -282,6 +282,28 @@ class Roaring {
         return api::roaring_bitmap_remove_range_closed(&roaring, min, max);
     }
 
+        /**
+     * Keep only values in the half-open interval [min, max).
+     * Equivalent to two consecutive removeRange calls.
+     */
+    void mask(uint64_t min, uint64_t max) noexcept {
+        removeRange(0, min);
+        if (!isEmpty()) {
+            removeRange(max, (uint64_t)maximum() + 1);
+        }
+    }
+
+    /**
+     * Keep only values in the closed interval [min, max].
+     * Equivalent to two consecutive removeRangeClosed calls.
+     */
+    void maskClosed(uint32_t min, uint32_t max) noexcept {
+        if (min > 0) removeRangeClosed(0, min - 1);
+        if (!isEmpty() && max < maximum()) {
+            removeRangeClosed(max + 1, maximum());
+        }
+    }
+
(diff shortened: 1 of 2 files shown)


lemire:
I like it well enough.

ping @fbbdev

SylvesterKwon:
Nice, this looks handy!

Small thing is: #419 also wanted the range-restricted add/flip/andnot with faster & specialized implementation that would be more efficient compared to temp-mask-based approach, and i guess that's out-of-scope of this PR.

So IMHO, this should not close #419.

Yokiatch:
@SylvesterKwon i've updated PR from closes to Related to #419

fbbdev:
Hey there!

Thanks for this, I think these convenience methods are good to have as a starting point, even if they are indirect implementations (not sure if there would be anything to gain here with ad-hoc impls).

@SylvesterKwon agreed the issue should not be closed as the range-restricted are probably the area that could gain the most for small ranges and large bitmaps.

lemire:
Just a reminder that there is no point in doing performance optimization unless there are realistic benchmarks that we care about. And even with such benchmarks, we need a reference optimization (so that we can justify the performance gains).

Thus I am not preoccupied by the fact that this PR is basically just providing the functionality with our existing C API (without extending it). It is a proper starting point.

lemire:
Merging.

PR #818, opened 2026-05-24
Wasm SIMD128 beyond bitsets + amalgamation wasm differential harness

### LINKED ISSUE #801: WASM SIMD128 backend
**Is your feature request related to a problem? Please describe.**

When CRoaring is compiled for `wasm32` (e.g. `wasm32-wasi` for browser / Node consumption), every container loop falls back to scalar — the x86 (`CROARING_IS_X64`) and ARM (`__ARM_NEON`) SIMD paths in `include/roaring/portability.h` do not fire, and there is no WASM SIMD128 equivalent. Browser-deployed CRoaring therefore runs the same bitset AND / OR / XOR / ANDNOT and popcount inner loops as a native scalar build, even though WASM SIMD128 has been broadly shipped since 2022 (Chrome 91+, Firefox 122+, Safari 16.4+).

We've confirmed this is not something Clang's auto-vectorizer recovers from. Compiling CRoaring v4.6.1 with `-msimd128` and inspecting the output with `wasm-objdump -d`:

```
# at -Os (36,297 bytes):
$ wasm-objdump -d facetsearch-core.wasm \
    | grep -oE 'v128\.[a-z0-9._]+|[fi][0-9]+x[0-9]+\.[a-z0-9._]+' | wc -l
0

# at -O3 (846,074 bytes):
0
```

Zero v128 opcodes at either level. The container loops are too branch-heavy for scalar-evolution vectorization; a hand-written SIMD backend is required to close the gap.

**Describe the solution you'd like**

A third arch branch in `portability.h` alongside the existing two:

```c
#if defined(__wasm_simd128__)
#define CROARING_IS_WASM 1
#include <wasm_simd128.h>
#endif
```

…and WASM-SIMD versions of the container-internal hot loops that already have SSE2 counterparts. SSE2 → WASM SIMD128 maps directly for most of the ops CRoaring uses:

| SSE2 | WASM SIMD128 |
|---|---|
| `_mm_and_si128`    | `wasm_v128_and` |
| `_mm_or_si128`     | `wasm_v128_or` |
| `_mm_xor_si128`    | `wasm_v128_xor` |
| `_mm_andnot_si128` | `wasm_v128_andnot` |
| `_mm_loadu_si128`  | `wasm_v128_load` |
| `_mm_storeu_si128` | `wasm_v128_store` |
| `__builtin_popcountll` × 2 | `wasm_i64x2_popcnt` (v128 popcount is 8-bit; bitset popcount needs a short reduce) |

**Rationale / expected objective benefit**

The highest-leverage targets are bitset AND / OR / XOR / ANDNOT in `src/containers/bitset.c` and the bitset cardinality popcount. Based on the SSE2 vs scalar delta on the same loops on native builds, we'd expect roughly 2–4× on bitset-heavy workloads once WASM SIMD128 is used — exactly the speedup x86/ARM users already
get.

Expected concrete wins for browser CRoaring consumers:

- **Bitset intersection / union / xor / andnot** — 4× 64-bit lanes per 128-bit op; currently 1-lane scalar.
- **Bitset cardinality (`roaring_cardinality`, `intersect_cardinality`, `union_cardinality`)** — `wasm_i64x2_popcnt` vectorizes the hot popcount loop.
- **Array container intersection / union (sorted u16)** — `wasm_i16x8_*` min/max/eq for the galloping-search and shuffle- merge steps. Matches the SSE2 implementation in `src/containers/array.c`.

**Describe alternatives you've considered**

1. **Enabling `-msimd128` and relying on Clang's auto-vectorizer.** Measured above: zero v128 opcodes produced, at any optimization level. Auto-vectorization is not a viable alternative for the container loops.

2. **SIMDe shim layer** (`simde/wasm/simd128.h` for SSE2-style intrinsics, or `simde/x86/sse2.h` mapped to WASM SIMD when `__wasm_simd128__` is set). Would in principle let the existing SSE2 code paths compile on WASM with minimal changes, but adds a header-only dependency and its code-gen on the bitset popcount path is known to produce slightly worse shuffles than hand-written `wasm_i64x2_popcnt`. Mentioning it for completeness — it's the way to unblock this without a full per-op port.

**Additional context**

- Tested CRoaring version: v4.6.1 (tarball sha256 `dd5c73c464be2df05b3472cce00d2d753b7eed0856752d88b23058aaff542216`).
- Toolchain: Zig 0.13.0 vendoring Clang 17, target `wasm32-wasi`, reactor exec model.
- Runtime support: browsers with WASM SIMD128 today cover >95% of global traffic. No runtime dispatch is needed if a consumer ships two builds (SIMD + scalar), same model as x86 SSE2 vs AVX2; or the SIMD build can simply be the baseline for evergreen-browser-only projects.

**Are you willing to contribute code or documentation toward this new feature?**

Yes

stereobooster:
## Summary

Extend Wasm SIMD128 beyond `src/containers/bitset.c` to the rest of the hot
paths, wire it into array / bitset / run conversion code, and add a
differential-digest harness that proves native, scalar-wasm, and `-msimd128`
wasm produce bit-identical output.

Closes #801. Primarily a perf change.

### SIMD coverage

- `array_util.c` — Schlegel/Lemire-style intersect / union / xor / difference,
  optional `memequals`, uint16→uint32 widen, and a new
  `intersect_vector16_nonempty` for balanced `array_container_intersect` parity
  with the pcmpestra/shuffle-mask paths. SIMD vs scalar speed for the Nonempty
  variant is workload-dependent; correctness is tied to the differential
  digests, not to `tools/wasm_simd_perf_smoke.sh` (whole-harness wall clock).
- `bitset_util.c`, `bitset.c` — bitset merges and SIMD bit-extract.
- `run.c` — `run_container_to_uint32_array` uses chunked `i32x4` widening
  (mirroring the AVX2 path) under `__wasm_simd128__` / `CROARING_WASM_SIMD`,
  with no separate opt-in switch.
- `convert.c` / `array.c` — `array_container_from_bitset` calls
  `bitset_extract_setbits_wasm_uint16` when SIMD is enabled and the array
  capacity clears the SIMD headroom guard; otherwise scalar.

Wasm-specific helpers use a `croaring_wasm_*` prefix (including
`croaring_wasm_array_container_to_uint32_array`) to avoid colliding with the
`wasm_*` intrinsics in `wasm_simd128.h`.

### Verification

- `tests/wasm_diff_harness.c` — amalgamation workload that compares
  deterministic digests across native, Emscripten scalar wasm, and
  `emcc -msimd128`. Guarded by SIMD-uplift checks (`wasm-objdump` heuristic on
  the linked wasm + roaring-only objects) and an amalgamation LLVM IR trace
  for `@llvm.wasm.*` / `@llvm.ctpop.v16i8`. Temporaries are freed so leak
  sanitizers stay quiet.
- `tools/run_wasm_differential_test.sh` drives the harness end-to-end.

### CI / tooling

- `.github/workflows/emscripten.yml` runs on `ubuntu-latest` via `setup-emsdk`
  (no Docker), and pipes through `tools/check_wasm_simd_neon_pairing.sh`
  (structural heuristic on Wasm/NEON `elif` hygiene) and
  `tools/ci_wasm_differential_digest.sh`.
- CMake convenience targets: `wasm_differential_test`,
  `wasm_neon_simd_parity_check`.
- `tools/run-wasm-diff-docker.sh` — optional local runner that builds a pinned
  Emscripten 3.1.74 + `wabt` image from an inline Dockerfile. CI does not use
  this path.
- `tools/wasm_simd_perf_smoke.sh` + `tools/_wasm_perf_bench_helper.py` —
  optional throughput smoke. Wall-clock noise-prone; complements but does not
  replace the digest correctness story.

### Misc

- `.gitignore` `build-debug/` and `.DS_Store`.


stereobooster:
Okay this is huge one. I can try to split it.

lemire:
Please be careful with unmotivated changes.

Capture d’écran, le 2026-05-28 à 21 39 14

stereobooster:
> Please be careful with unmotivated changes.

yes this needs a break down. It is not reviewable in current shape.

Do you have any high-level feedback, regarding test-strategy choice, for example. Do not look at small details just yet

stereobooster:
Closing this in favor of a set of small, single-purpose PRs

**Open now**
- #835 — Zig translate-c local rename (independent cleanup)
- #836 — native metamorphic correctness tests (independent of wasm)
- #837 — WASM SIMD128 differential test harness + structural linter (verification infra; lands first)

**Coming next, stacked on #837 (opened as each predecessor merges, to keep diffs clean and CI green):**
1. `run_container_to_uint32_array` + portability foundation (`CROARING_WASM_SIMD`)
2. bitset containers + `array_container_from_bitset`
3. array intersect/union/xor/difference
4. `intersect_vector16_nonempty`

**Dropped from the original branch as unrelated/unmotivated:** the `.DS_Store`/`build-debug` gitignore churn, a stray `roaring.hh` whitespace edit, and the perf/Docker convenience tooling (`wasm_simd_perf_smoke.sh`, `_wasm_perf_bench_helper.py`, `run-wasm-diff-docker.sh`) — these were perf/convenience, not correctness, and can come later if wanted.

The differential-digest CI gate is added together with the first SIMD path so every PR stays green.

Issue #819, opened 2026-05-24
wasm32-wasi: roaring_bitmap_to_uint32_array output can disagree with documented sorted-array contract


stereobooster:
**Describe the bug**

With CRoaring v4.6.1 built for wasm32-wasi (see configuration), `roaring_bitmap_to_uint32_array` can fill the output buffer with `n = roaring_bitmap_get_cardinality(r)` values that are not the sorted unique list of integers in the bitmap. In one case the bitmap should be `{0 … 49 999}` (cardinality 50 000), but the array contained only 49 992 distinct values (duplicates), while cardinality still reported 50 000. Portable serialize/deserialize round-trip still looked consistent with `r` in downstream checks.

The v4.6.1 header documents that the function converts the bitmap to a sorted array, with the caller allocating `roaring_bitmap_get_cardinality(bitmap)` elements:

https://github.com/RoaringBitmap/CRoaring/blob/v4.6.1/include/roaring/roaring.h

The v4.6.1 README shows reading back with `roaring_bitmap_of_ptr(cardinality, arr)` and `roaring_bitmap_equals` after `roaring_bitmap_to_uint32_array`:

https://github.com/RoaringBitmap/CRoaring/blob/v4.6.1/README.md

A workaround that behaved correctly for the same test was iteration via `roaring_iterator_create` and `roaring_uint32_iterator_read`.

**To Reproduce**

1. Build CRoaring v4.6.1 for wasm32-wasi (here using Zig 0.13.0 to compile the C sources with `-DCROARING_DISABLE_X64`, `-DCROARING_DISABLE_NEON`, `-DCROARING_DISABLE_AVX512`, `-DDISABLE_X64`, and `CROARING_IS_BIG_ENDIAN=0`).
2. Create a bitmap and add every integer from 0 to n−1 for n = 50 000.
3. Check `roaring_bitmap_get_cardinality(r) == n`.
4. Allocate `uint32_t ans[n]` and call `roaring_bitmap_to_uint32_array(r, ans)`.
5. Count distinct values in `ans` (e.g. after sort, or with a set). Observed: fewer than n distinct values despite cardinality n (we saw 49 992 for n = 50 000).

Optional: round-trip with `roaring_bitmap_portable_serialize` / `roaring_bitmap_portable_deserialize_safe` and compare cardinality or `roaring_bitmap_equals`.

[...]

PR #820, opened 2026-05-24
Fix wasm32 to_uint32_array correctness (issue #819)

stereobooster:
## Summary

- Adds regression test **`issue819`**: dense bitmap `{0 … 49999}` must match iterator bulk reads from `roaring_bitmap_to_uint32_array` ([issue #819](https://github.com/RoaringBitmap/CRoaring/issues/819)).
- Replaces `memcpy`-based scalar writes with plain `uint32_t` assignments in bitset/array/run exporters so wasm32 backends do not corrupt the exported array output.

## Test plan

- [x] `cmake --build build --target toplevel_unit && ./tests/toplevel_unit` (includes `issue819`)
- [x] `tools/run-clangcldocker.sh` (skipped here; Docker not run in this environment)
- [ ] Rebuild wasm32-wasi consumer and confirm facetsearch/`toArray`-style scenario from the issue

Made with [Cursor](https://cursor.com)

### DIFF
--- a/src/containers/array.c
+++ b/src/containers/array.c
@@ -418,10 +418,7 @@ int array_container_to_uint32_array(void *vout, const array_container_t *cont,
     uint32_t *out = (uint32_t *)vout;
     size_t i = 0;
     for (; i < (size_t)cont->cardinality; ++i) {
-        const uint32_t val = base + cont->array[i];
-        memcpy(out + outpos, &val,
-               sizeof(uint32_t));  // should be compiled as a MOV on x64
-        outpos++;
+        out[outpos++] = base + cont->array[i];
     }
     return outpos;
 }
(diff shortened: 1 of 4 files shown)


src/containers/array.c
```diff
@@ -418,10 +418,7 @@ int array_container_to_uint32_array(void *vout, const array_container_t *cont,
     uint32_t *out = (uint32_t *)vout;
     size_t i = 0;
     for (; i < (size_t)cont->cardinality; ++i) {
-        const uint32_t val = base + cont->array[i];
-        memcpy(out + outpos, &val,
-               sizeof(uint32_t));  // should be compiled as a MOV on x64
-        outpos++;
```
Dr-Emann, line 424: Other than the old code handling un-aligned data, and the new code _not_ these should behave identically. Please explain why this would lead to an error in WASM.

Same to all other changes

stereobooster: yes you are totally right. I just realized this myself. When I tried to bisect root cause

PR #840, opened 2026-06-21
feat: Add Roaring64

### LINKED ISSUE #629: Rewrite the C++ Roaring64Map implementation to use 64-bit croaring

SylvesterKwon:
Closes #629.

It adds the new `Roaring64` class discussed in #629, a C++ wrapper over the ART-based 64-bit C impl. Existing 64-bit C++ impl `Roaring64Map` remains untouched.

This PR implements core ops only (for now). The rest are left as TODOs to keep the PR small.

Included: 

- construction
- add/addMany/remove/clear
- contains/cardinality/isEmpty/minimum/maximum/operator==
- and/or/andnot/xor (in-place + value-returning)
- a forward const iterator
- toArray

Tests:

- tests/cpp_roaring64_unit.cpp: unit tests for the above (set ops span multiple ART containers),
- tests/roaring64_checked.hh: doublechecked harness for `Roaring64` (as `Roaring64Map` has `roaring64map_checked`). Cross-checking a `Roaring64` against `std::set` over a randomized ops sequence.

Example:

```cpp
using roaring::Roaring64;

// construction & basic ops
Roaring64 a{1, 2, 100, uint64_t(1) << 40};   // initializer list
a.add(7);
a.contains(100);   // true
a.cardinality();   // 5
a.minimum();       // 1
a.maximum();       // 1 << 40

// set operations (value-returning and in-place)
Roaring64 b{2, 3, uint64_t(1) << 40};
Roaring64 inter = a & b;   // {2, 1<<40}
Roaring64 uni   = a | b;
a |= b;                    // in-place op

// iterator
Roaring64 r{5, 5, 7, uint64_t(1) << 40};

// 1. range-based for
for (uint64_t v : r) {
    // ...
} 

// 2. explicit begin()/end()
for (auto it = r.begin(); it != r.end(); ++it) {
    uint64_t v = *it;
}

// works with STL
std::vector<uint64_t> v(r.begin(), r.end());           // sorted copy
uint64_t sum = std::accumulate(r.begin(), r.end(), 0ull);
bool has7    = std::find(r.begin(), r.end(), 7) != r.end();
auto n       = std::distance(r.begin(), r.end());      // == r.cardinality()
```

Out of scope for this PR, planned as follow-up PRs:

- Add missing APIs. Such as...
    - serialization and frozen API
    - rank / select / getIndex
    - range ops
    - run_optimize / removeRunCompression / shrinkToFit
    - fastunion, statistics, printf
    - uint32_t overloads (do we need these?)
- Improve iterator
    - ~support backward iteration~ (DONE)
    - random access
- Improve roaring64_checked, since it only has minimal methods for current test
- Documentation. e.g., when to use `Roaring64Map` vs `Roaring64`?

(Also fixed a "materalization" typo in roaring.hh)


SylvesterKwon:
cc @madscientist, would appreciate your review. Thanks in advance.

cpp/roaring/roaring64.hh
```diff
@@ -0,0 +1,472 @@
...
+        if (roaring == nullptr) {
+            ROARING_TERMINATE("failed roaring64_bitmap_copy in constructor");
+        }
+    }
+
+    /**
+     * Move constructor. The moved-from bitmap is left in a null state and may
+     * only be destroyed or assigned to.
+     */
+    Roaring64(Roaring64&& r) noexcept : roaring(r.roaring) {
```
lemire, line 190: **[suggestion]** The move constructor sets `r.roaring = nullptr`, leaving the moved-from object in a null state. Unlike `Roaring` (which calls `roaring_bitmap_init_cleared` and remains a valid empty bitmap), most `Roaring64` methods (`isEmpty()`, `cardinality()`, `begin()`, `operator==`, copy ctor, etc.) will dereference the null pointer if invoked on a moved-from object. The comment documents "may only be destroyed or assigned to," but this diverges from the established `Roaring` API and is easy to misuse.

**Suggestion:** After the move, assign `r.roaring = api::roaring64_bitmap_create()` (and terminate on failure), or otherwise leave a valid empty bitmap so moved-from objects behave like a default-constructed instance—matching `Roaring`.

cpp/roaring/roaring64.hh
```diff
@@ -0,0 +1,472 @@
...
+     */
+    Roaring64(std::initializer_list<uint64_t> l) : Roaring64() {
+        addMany(l.size(), l.begin());
+    }
+
+    /**
+     * Construct a roaring object by taking control of a malloc()'d C struct.
+     * The pointer to the C struct will be invalid after the call.
+     */
+    explicit Roaring64(roaring64_bitmap_t* s) noexcept : roaring(s) {}
```
lemire, line 174: **[suggestion]** `explicit Roaring64(roaring64_bitmap_t* s)` takes ownership of a malloc'd C struct with no null check and no documentation that a null pointer is unsafe. The 32-bit `Roaring(roaring_bitmap_t* s)` constructor explicitly warns "Passing a NULL pointer is unsafe."

**Suggestion:** Add the same warning to the doc comment (and optionally `ROARING_TERMINATE` or assert on `s == nullptr`) for parity with `Roaring`.

cpp/roaring/roaring64.hh
```diff
@@ -0,0 +1,472 @@
...
+class Roaring64ConstIterator {
+   public:
+    typedef std::forward_iterator_tag iterator_category;
+    typedef uint64_t value_type;
+    typedef std::ptrdiff_t difference_type;
+    typedef void pointer;
+    typedef uint64_t reference;
+
+    /** Begin iterator over roaring64_bitmap_t. */
+    explicit Roaring64ConstIterator(const api::roaring64_bitmap_t* r)
```
lemire: **[suggestion]** `Roaring64ConstIterator` calls `roaring64_iterator_create(r)` in its constructor without checking the result. If `roaring_malloc` fails, `roaring64_iterator_create` passes a null pointer into `roaring64_iterator_init_at`, which writes through it (undefined behavior). On success with an empty bitmap, a failed allocation would be indistinguishable from a legitimate end iterator only after a crash.

**Suggestion:** Check the return value of `roaring64_iterator_create` (and `roaring64_iterator_copy`) and call `ROARING_TERMINATE` on failure, consistent with other allocating paths in this header.

cpp/roaring/roaring64.hh
```diff
@@ -0,0 +1,472 @@
...
+
+    /**
+     * Clears the bitmap.
+     */
+    void clear() noexcept { api::roaring64_bitmap_clear(roaring); }
+
+    /**
+     * Return the largest value (if not empty)
+     */
+    uint64_t maximum() const noexcept {
```
lemire, line 282: **[suggestion]** `minimum()` and `maximum()` are documented as returning the smallest/largest value "(if not empty)" but do not describe the C API sentinel values used when empty (`UINT64_MAX` for minimum, `0` for maximum per `roaring64.h`). Callers who skip an `isEmpty()` check may misinterpret results.

**Suggestion:** Document the empty-bitmap return values in the method comments (as the C header does), or add a brief class-level note. Consider a test in `cpp_roaring64_unit.cpp` asserting this behavior.

cpp/roaring/roaring64.hh
```diff
@@ -0,0 +1,472 @@
...
+    Roaring64ConstIterator operator++(int) {
+        Roaring64ConstIterator orig(*this);
+        ++(*this);
+        return orig;
+    }
+
+    // TODO: operator-- for bidirectional iteration
+
+    /** True if both iterators are at the same position. */
+    bool operator==(const Roaring64ConstIterator& o) const {
```
lemire, line 118: **[nit]** `Roaring64ConstIterator::operator==` compares iterators by current value (`roaring64_iterator_value`) rather than iterator position. This is correct for same-bitmap forward iteration (values are unique) and handles end iterators via `atEnd()`, but two iterators from different bitmaps at the same numeric value would compare equal, which is misleading for invalid cross-bitmap comparisons.

**Suggestion:** Add a brief comment that equality is intended for comparing iterators within the same bitmap iteration loop, or compare underlying iterator state if a C helper becomes available.

cpp/roaring/roaring64.hh
```diff
@@ -0,0 +1,472 @@
...
+    ~Roaring64() {
+        if (roaring != nullptr) {
+            api::roaring64_bitmap_free(roaring);
+        }
+    }
+
+    /**
+     * Construct a bitmap from a list of uint64_t values.
+     */
+    static Roaring64 bitmapOf(size_t n, ...) {
```
lemire, line 234: **[suggestion]** `Roaring64` provides `bitmapOf(size_t n, ...)` and an initializer-list constructor but lacks `bitmapOfList(std::initializer_list<uint64_t>)`, which both `Roaring` and `Roaring64Map` expose. Users may expect API symmetry across the three wrapper classes.

**Suggestion:** Add `static Roaring64 bitmapOfList(std::initializer_list<uint64_t> l)` delegating to `addMany`, mirroring `Roaring64Map::bitmapOfList`.

tests/cpp_roaring64_unit.cpp
```diff
@@ -0,0 +1,253 @@
...
+    const uint64_t seedvals[] = {1, uint64_t(1) << 40, uint64_t(1) << 50,
+                                 12345};
+    r.addMany(4, seedvals);
+    for (uint64_t v : seedvals) added.push_back(v);
+
+    for (size_t i = 0; i < 5000; ++i) {
+        if (!added.empty() && (next() & 3) == 0) {  // 25%
+            // Remove an existing value.
+            uint64_t v = added[next() % added.size()];
+            r.remove(v);
```
lemire, line 264: **[nit]** In `test_cpp_r64_random_vs_set`, the remove branch calls `r.remove(v)` but does not remove `v` from the local `added` vector. The test still passes because `doublechecked::Roaring64` keeps `plain` and `check` in sync and removing an absent value is a no-op, but `added` can accumulate stale entries and skew the random remove/contains distribution.

**Suggestion:** Erase `v` from `added` after a successful remove (e.g., find and erase one matching entry) so the shadow vector stays accurate.

tests/cpp_roaring64_unit.cpp
```diff
@@ -0,0 +1,253 @@
...
+
+#include "roaring64_checked.hh"
+#include "test.h"
+
+using roaring::Roaring64;
+
+static_assert(std::is_nothrow_move_constructible<Roaring64>::value,
+              "Expected Roaring64 to be nothrow move constructible");
+
+DEFINE_TEST(test_cpp_r64_default_empty) {
```
lemire, line 20: **[suggestion]** Tests cover default empty, copy/move, set ops, and iteration well, but do not exercise empty-bitmap `minimum()`/`maximum()` sentinel behavior, self-assignment (`c = c`), or `operator-=` / `operator^=` with `&r == this` (documented UB). The latter is acceptable to omit, but empty min/max is worth locking in given the ambiguous C++ doc wording.

**Suggestion:** Add a small test asserting `minimum()`/`maximum()` on an empty bitmap match the C API contract, and optionally that `a = a` preserves contents.

cpp/roaring/roaring64.hh
```diff
@@ -0,0 +1,472 @@
...
+
+    /**
+     * Return the largest value (if not empty)
+     */
+    uint64_t maximum() const noexcept {
+        return api::roaring64_bitmap_maximum(roaring);
+    }
+
+    /**
+     * Return the smallest value (if not empty)
```
lemire: Document what happens if empty.

lemire (commented):
## Summary

This PR adds a well-structured C++ `roaring::Roaring64` wrapper over the ART-based `roaring64` C API, with build/install/amalgamation wiring, unit tests, and a `doublechecked::Roaring64` test helper. The implementation closely mirrors existing `Roaring` / `Roaring64Map` patterns: RAII ownership, terminating allocators for materializing operators, and solid coverage of core add/remove/iterate/set-op semantics. The dominant risks are (1) moved-from objects holding a null internal pointer—documented but unlike `Roaring`, which leaves a valid empty bitmap—and (2) iterator/ownership edge cases (OOM on iterator allocation, null pointer constructor) that are not guarded or documented as thoroughly as the 32-bit wrapper.
## Issue counts by severity

- bugs: 0
- suggestions: 6
- nits: 2

cpp/roaring/roaring64.hh
```diff
@@ -0,0 +1,472 @@
...
+
+    // TODO: Implement removeRangeClosed(uint64_t min, uint64_t max)
+
+    /**
+     * Clears the bitmap.
+     */
+    void clear() noexcept { api::roaring64_bitmap_clear(roaring); }
+
+    /**
+     * Return the largest value (if not empty)
```
lemire: Document what happens if empty.

lemire:
@SylvesterKwon Review above is by AI (except for my comments that maximum/minimum should be documented). Take with a grain of salt. I have not personally reviewed it yet. Please look at the AI review.

cpp/roaring/roaring.hh
```diff
@@ -820,7 +820,7 @@ class Roaring {
     Roaring operator|(const Roaring &o) const {
         roaring_bitmap_t *r = api::roaring_bitmap_or(&roaring, &o.roaring);
         if (r == NULL) {
-            ROARING_TERMINATE("failed materalization in or");
+            ROARING_TERMINATE("failed materialization in or");
```
lemire, line 823: Appreciated.

cpp/roaring/roaring64.hh
```diff
@@ -0,0 +1,472 @@
+/**
+ * A C++ header for ART-based 64-bit Roaring Bitmaps.
+ *
+ * Reference (format specification) :
+ * https://github.com/RoaringBitmap/RoaringFormatSpec#extention-for-64-bit-implementations
```
lemire, line 5: good

cpp/roaring/roaring64.hh
```diff
@@ -0,0 +1,472 @@
...
+        va_end(vl);
+        return ans;
+    }
+
+    /**
+     * Adds value x.
+     */
+    void add(uint64_t x) noexcept { api::roaring64_bitmap_add(roaring, x); }
+
+    // TODO: Implement addChecked(uint64_t x)
```
lemire: I'd prefer if you did not leave TODOs but instead opened issues that we can complete later.

cpp/roaring/roaring64.hh
```diff
@@ -0,0 +1,472 @@
...
+    }
+
+    /**
+     * Removes value x.
+     */
+    void remove(uint64_t x) noexcept {
+        api::roaring64_bitmap_remove(roaring, x);
+    }
+
+    // TODO: Implement removeChecked(uint64_t x)
```
lemire: Same comment here, it is fine not to implement everything, but leaving comments is distracting. Better documenting missing functionality elsewhere

lemire:
Ok. Manually reviewed. This looks fine to me. See my comments.

We'll need to update the README to document the usage, but it can be done either now or in a subsequent PR/commit. It is fine.

cpp/roaring/roaring64.hh
SylvesterKwon: Removed it. The "position" wording was inaccurate, especially at the iterator at the end. Thanks.

cpp/roaring/roaring64.hh
SylvesterKwon: Added. [feat: Add Roaring64::bitmapOfList](https://github.com/RoaringBitmap/CRoaring/pull/840/commits/56cd335b8bfabc4b746f85fd2cc60518d10663c7)
Since it's just simple delegation to `addMany`, no need to leave as a TODO.

tests/cpp_roaring64_unit.cpp
SylvesterKwon: Added the min/max for empty set test but I'll skip the documented UB part.

cpp/roaring/roaring64.hh
SylvesterKwon: Fixed. [docs: Document empty min/max return](https://github.com/RoaringBitmap/CRoaring/pull/840/commits/e09f44bd7d53c976871c5a7f9632cbc78e3a5c59)

I found out the same issue existed in both 32-bit implementation and Roaring64Map. So I fixed them together in this commit.

cpp/roaring/roaring64.hh
SylvesterKwon: Done. [docs: Warn about NULL in Roaring64 pointer ctor](https://github.com/RoaringBitmap/CRoaring/pull/840/commits/287f92f53def2b45a82a30c37b4cc87025cf9017)

Added the doc warning but no runtime check.

tests/cpp_roaring64_unit.cpp
SylvesterKwon: Ok this was obvious mistake. Fixed. [fix: test_cpp_r64_random_vs_set](https://github.com/RoaringBitmap/CRoaring/pull/840/commits/4a3b3c0df6fee3555af91be3d1d8f9de2cb116dc)

cpp/roaring/roaring64.hh
SylvesterKwon: In my understanding, this is being addressed in #838. So I'll leave the wrapper as-is for now. I'll align this after the memory policy is settled.

cpp/roaring/roaring64.hh
SylvesterKwon: Leaving as-is -- making the moved-from state a valid empty bitmap would need an allocation for it, which breaks the noexcept move (`roaring64_bitmap_create`). And we already described that only destroy/assign are allowed for the null state.

cpp/roaring/roaring64.hh
SylvesterKwon: Removed the inline TODOs. [misc: Remove unimplemented-method TODO comments](https://github.com/RoaringBitmap/CRoaring/pull/840/commits/0cadc3e6e4df2dd26b27c823450872a4131a0289)

SylvesterKwon:
Thanks for taking the time to review this. I checked both the AI and manual review and replied to each item.

On documentation -- I plan to document once the API is a bit more complete (serialization especially).

I'll open follow-up issues for the remaining Roaring64 functionality once this is merged, and work through them on weekends.

cpp/roaring/roaring64.hh
```diff
@@ -0,0 +1,420 @@
...
+
+namespace roaring {
+
+/**
+ * Const iterator of a Roaring64.
+ * Wraps a roaring64_iterator_t. An end iterator holds a null handle.
+ */
+class Roaring64ConstIterator {
+   public:
+    typedef std::forward_iterator_tag iterator_category;
```
Dr-Emann: On my phone, so can't easily check, but I think we should be able to be a bidirectional iterator?

lemire: Looks like a fair point: @SylvesterKwon do you want to cover this?

SylvesterKwon: Yes! I'm currently working on this.

SylvesterKwon: Done. [feat: Make Roaring64 iterator bidirectional](https://github.com/RoaringBitmap/CRoaring/pull/840/commits/cf0749ba4dcb19bed09f2d3439e420e4b414eb68)

Iterator is now bidirectional. Since `roaring64_iterator_t` is opaque, I keep `end()` as a cheap sentinel and lazily materialize its handle when `--end()` is called.

Take a look and let me know if anything needs improving. Thanks!

SylvesterKwon:
@lemire Gentle reminder on this PR

lemire:
@SylvesterKwon Thanks. I am ok with merging this PR. Before I do, if you want, you could add some documentation in the README.md. If you don't, I will do so after merging this PR.

SylvesterKwon:
> @SylvesterKwon Thanks. I am ok with merging this PR. Before I do, if you want, you could add some documentation in the README.md. If you don't, I will do so after merging this PR.

Done. [docs: Document Roaring64 in the README](https://github.com/RoaringBitmap/CRoaring/pull/840/commits/724d584355a51208a932f0a9112f8936fe5bb626)

I listed `Roaring64` alongside other C++ classes. And I've marked it as experimental for now since it lacks some functionality for this scope.

Let me know if anything's worth expanding.

PR #848, opened 2026-07-16
Extract roaring_array_t to `ra` variables

AE1020:
> *(Note: this commit is not intended to change any functionality.  It's
intended to clarify code so that some related PRs will be easier to
audit, vs trying to mix the refactoring and the material changes in
the same commit.  AI should be able to verify that no semantic
difference was made by these edits.)*

---

In roaring.c in particular, there was significant repetition of access
of `high_low_container`, with code like:

    new_container = container_remove(
        r->high_low_container.containers[pos], vals[i] & 0xffff,
        r->high_low_container.typecodes[pos], &new_typecode);
     if (new_container != r->high_low_container.containers[pos]) {
        container_free(r->high_low_container.containers[pos],
                       r->high_low_container.typecodes[pos]);
        ra_replace_key_and_container_at_index(&r->high_low_container,
                                              pos, key, new_container,
                                              new_typecode);

This repetition was a significant tax on comprehension, and with
a simple definition like:

    roaring_array_t *ra = &r->high_low_container;

It becomes instantly more understandable, e.g.:

    new_container = container_remove(
        ra->containers[pos], vals[i] & 0xffff,
        ra->typecodes[pos], &new_typecode);
    if (new_container != ra->containers[pos]) {
        container_free(ra->containers[pos],
                       ra->typecodes[pos]);
        ra_replace_key_and_container_at_index(
             ra, pos, key, new_container, new_typecode);

Some routines had done this factoring, while others had not.  This
goes through the files that use `high_low_container` more than once
in a function, and factors out a local variable to hold the address.

In optimized builds, this should generate the same code.

In order to make the naming systemic with `ra` chosen for the stem
of the variable, the various alernative names for roaring_bitmap_t
code (`bm`, `x`) are synchronized to being based on `r`.  The plain
name `r` is then taken for a roaring bitmap that is going to be
given back as a result (as opposed to `ans` or `answer`), with numbers
used to disambiguate input arguments of type roaring_bitmap_t.  This
allows `ra1` to correspond to `r1`, `ra2` to correspond to `r2`, and
plain `ra` to correspond to `r` as the generated result.


lemire:
For the time being, I manually review the PRs (reading each line). Here it seems that you use AI to refactor hundreds of lines of code. The sensible way to review this would be for me to use AI. But if AI is doing the reviews, then why do we care about readability?


Our PR template says the following:


> We welcome most contributions. However, most contributions should fall in our of these three categories:
> Bug fixes. If you found a new bug, please include a corresponding test.
> New features.
> Better performance. Provide benchmark results.


This PR does not fall in this scope. 

You are renaming local variables...

Capture d’écran, le 2026-07-16 à 19 11 06


I am not going to review this. It is too long. It is arbitrary.

AE1020:
I did not use AI to do this.

You--however can review it using AI to ask if it is identity/semantics preserving.

Further PRs are easier to understand with this.  I can send you other PRs without it, however I put it first because I did not want to combine semantics changes with identity-preserving refactorings.

AE1020:
It is--certainly possible for me to send you the "value-adding" PRs first.  If that's what you insist on, that is fine.  These changes can be reordered later.  I was trying to make things easier to audit, by giving you something that could be checked as semantics-preserving.

Perhaps if you see some value first you will be more inclined to understand the goals.  Again, that is fine; I will reorder the commits.  My intent was to make auditing easier--not harder.

lemire:
@AE1020 Please don't refactor the library with or without AI.

I am not going to invest time reviewing hundreds of line of code because you prefer one variable name rather than some other variable name. 

We are not doing this. I don't have time for this. Accepting such PRs is likely to introduce bugs in otherwise battle tested code.

I am not replacing code that ran in production for years just so someone finds it 'more understandable'. No. Just no.

lemire:
@AE1020 

PRs should...

1. Fix a bug.
2. Improve performance (with a benchmark).
3. Add a feature.

We make it explicit in the PR template that we are not inviting the 'let me rename the variables because I like it better that way' kind of PRs. No. No. No.

PR #849, opened 2026-07-17
Build-C-as-C++ Patches

AE1020:
As per [C++ core guidelines CPL.2](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#rcpl-subset), it's best for C code to be buildable with a C++ compiler.  *(my codebase using CRoaring has to build as both C99+ and C++11+)*.  Because of this I had contributed some of the necessary changes to CRoaring to achieve that some years ago.

At that time, I also slipped in some "experimental" ideas involving how `container_t*` could be a `void*` in C, but in C++ it could use the "empty base class optimization" with the container subclasses inheriting from it.  This meant it would be type checked more strongly if built as C++...so that you got the benefit of typed inheritance...only allowing containers to be passed to `container_t*` vs arbitrary pointers.

It was the *start* of a good idea, but only the start *(and the way I did it then was often more clutter than it provided benefit, as well as an ABI break with C-built code in some compilers in some circumstances)*.  Those ideas have evolved into a **vastly** richer framework now... which may be applied to many problems.  Including solving your [memory allocation failures/propagation](https://github.com/RoaringBitmap/CRoaring/issues/630) in a "Rust-like" way, by using the type system to annotate the return values of routines that can fail, and forcing compile-time errors when triage is not performed:

  https://needful.metaeducation.com/

This new model removes the wacky C++ from CRoaring's codebase entirely; the "enhancements" are in a repository that only those who wish to build with instrumentation "install".  It becomes a tool that you sideload in certain developer/CI builds--not part of CRoaring itself.  *(Your focus on "amalgamation" demonstrates you are interested in pragmatic packaging, and Needful is also very pragmatic: removing the clutter, while delivering the benefit.)*

Which is good: e.g. in my fork I've eliminated of all the below kind of junk, while delivering the actual value that a "C++ enhanced build" was theoretically supposed to provide:

https://github.com/RoaringBitmap/CRoaring/blob/d60ba3f447341a088bc8041277f6017f65787261/include/roaring/containers/container_defs.h#L73-L88

https://github.com/RoaringBitmap/CRoaring/blob/d60ba3f447341a088bc8041277f6017f65787261/include/roaring/containers/array.h#L54-L56

(etc.)  This is all gone, replaced by a ([via a single `downcast` operator](https://needful.metaeducation.com/cast#downcast-allows-mutable-source-to-const-target)), that's just `#define downcast (void*)` in the baseline C distribution (and [minor magic](https://github.com/metaeducation/needful/blob/f18e92891e656bf9fbb26e800eaffed34ebcb49b/needful.h#L160-L217) in non-enhanced C++ builds)... but [metaprogramming sorcery](https://github.com/metaeducation/needful-enhanced/blob/main/needful-casts.hpp) once you sideload the C++ enhanced files.  It detects the target type implicitly without you having to say what it is and validates the inheritance relationship, it won't downcast const pointers to non-const, it won't cast double-pointers to single ones, it can compile-time restrict the allowed to and from casts, or runtime validate the bit patterns based on what you're casting to what, and so on and so forth...

---

But to get the ball rolling to show you what I mean, the atrophy in C++ builds has to be fixed first.  Here are some (I hope) uncontroversial patches to start with.

### DIFF
--- a/src/bitset.c
+++ b/src/bitset.c
@@ -11,7 +11,7 @@
 #ifdef __cplusplus
 extern "C" {
 namespace roaring {
-namespace internal {
+namespace api {  // bitset_t was once internal, but made public in the API
 #endif
 
 extern inline void bitset_print(const bitset_t *b);
--- a/tests/c_example1.c
+++ b/tests/c_example1.c
@@ -7,6 +7,11 @@
 
 #include "test.h"
 
+#ifdef __cplusplus  // stronger type checking errors if C built in C++ mode
+using namespace roaring::api;
+using namespace roaring::misc;
+#endif
+
 bool roaring_iterator_sumall(uint32_t value, void *param) {
     *(uint32_t *)param += value;
     return true;  // iterate till the end
--- a/include/roaring/roaring_types.h
+++ b/include/roaring/roaring_types.h
@@ -19,7 +19,7 @@
 #ifdef __cplusplus
 extern "C" {
 namespace roaring {
-namespace api {
+namespace internal {
 #endif
 
 /**
@@ -38,13 +38,19 @@ namespace api {
 extern "C++" {
 struct container_s {};
 }
-#define ROARING_CONTAINER_T ::roaring::api::container_s
+#define ROARING_CONTAINER_T ::roaring::internal::container_s
 #else
 #define ROARING_CONTAINER_T void  // no compile-time checking
 #endif
 
 #define ROARING_FLAG_COW UINT8_C(0x1)
 #define ROARING_FLAG_FROZEN UINT8_C(0x2)
+
+#ifdef __cplusplus
+}  // end namespace internal (but still in namespace roaring)
+using internal::container_s;  // use scoped to this namespace block
+namespace api {                        // begin API namespace
+#endif
 
 /**
  * Roaring arrays are array-based key-value pairs having containers as values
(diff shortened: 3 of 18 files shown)


tests/c_example1.c
```diff
@@ -7,6 +7,11 @@
 
 #include "test.h"
 
+#ifdef __cplusplus  // stronger type checking errors if C built in C++ mode
+using namespace roaring::api;
```
AE1020, line 11: Upon review I notice this line is actually superfluous: including `roaring.h` (as opposed to `roaring.hh`) when building as `__cplusplus`) automatically includes the `roaring::api` namespace.

https://github.com/RoaringBitmap/CRoaring/blob/master/include/roaring/roaring.h#L1367-L1382

Perhaps that points to a potential separation of `configreport.h` vs. `configreport.hh` that does a similar trick; you don't need a `using namespace roaring::misc;` if you use the .h version?

src/bitset.c
```diff
@@ -11,7 +11,7 @@
 #ifdef __cplusplus
 extern "C" {
 namespace roaring {
-namespace internal {
+namespace api {  // bitset_t was once internal, but made public in the API
```
madscientist, line 14: I don't think this comment belongs in the code.  That's information for the Git commit message.

PR #851, opened 2026-08-01
Speed up array container union: remove qsort from the SIMD tail, add an AVX-512 bitonic merge

lemire:
Array container unions were the slowest of the array kernels by a wide margin, and neither of the two reasons was the one I expected. This fixes both.

Measurements are on a Xeon Gold 6548N (Emerald Rapids, `avx512vbmi2` / `vpopcntdq` / `bitalg`), glibc 2.39, gcc 14.3.

### Background: where the time actually goes

Cost per input element of the balanced array-container kernels on master:

| kernel | cycles/element |
| --- | --- |
| `intersect_vector16` | 0.50 |
| `difference_vector16` | 0.44 |
| **`union_vector16`** | **2.28** |
| **`xor_vector16`** | **2.51** |

Intersection is essentially at the hardware limit — `PCMPISTRM` retires 64 comparisons in ~3 cycles, and any AVX-512 all-pairs replacement needs 32 `vpermw` + 32 `vpcmpw`, all of which contend for port 5, costing ≥2 cycles/element. I prototyped and rejected two other AVX-512 candidates for the same reason: a 16-wide gather-based binary search for `intersect_skewed_uint16` (0.63–1.44x, gather throughput dominates) and a `vpconflictd`-based scatter for `bitset_set_list` (0.63–0.75x — the existing `bts` loop already runs at 1.45 cycles/element and beats scatter). Union and XOR were the real outliers.

### Commit 1: `qsort()` in the union/xor tails

`union_vector16()` and `xor_vector16()` finish by sorting a leftover buffer of **at most 16 `uint16` values**, and call `qsort()` to do it. On glibc that is a merge sort behind a function pointer, with its own temporary allocation and `memmove` traffic.

Profiling `union_vector16` on 64-element inputs attributes **53% of its cycles to that tail**: `msort_with_tmp` 23.8%, `memmove` 18.5%, `uint16_compare` 5.2%, `qsort_r` 2.4%.

The cost is fixed, ~80–130ns per call regardless of input length. That is enough to make `union_vector16` **slower than the scalar `union_uint16` it replaces at every size up to 2048 elements** — so on current master, dispatching to the SSE union is a pessimization on this machine except at the very largest container sizes.

An inline insertion sort over ≤16 values removes it. This one is portable and helps every target, not just x86.

### Commit 2: AVX-512 bitonic merge

`union_vector16` uses an 8-lane odd-even transposition network — 8 dependent min/max stages per 8 output values. That is O(n) stages per vector, so simply widening the registers buys nothing.

Replaced with a Batcher bitonic merge over 32 lanes: reverse one input, one min/max pair, then 5 compare-exchange stages per sorted half. Each distance has a cheap dedicated shuffle (`vshufi64x2`, `vpshufd`, `vprold`), so only the reverse needs a full `vpermw`. Dedup folds into one `vpcompressw` masked store against the previous lane. Result: ~1.3 cycles/element.

This is genuinely AVX-512-specific rather than a wider rerun of the SSE code — it needs cross-lane 16-bit permutes (`vpermw`, `vpermt2w`) and a 16-bit compress store (`vpcompressw`, AVX512_VBMI2), none of which exist in AVX2. Inputs below one 32-lane block fall through to `union_uint16`.

Kernel in isolation, n vs n, vs `union_vector16`: **8.6x** at n=32, 2.4x at 128, 2.0x at 512, 1.9x at 1024, 1.9x at 4096.

### End-to-end

`roaring_bitmap_or` over bitmaps whose containers are array containers of the given size (32 containers, min-of-15 rounds, pinned core, idle machine), ns:

| values/container | master | this PR | commit 1 alone | total |
| --- | --- | --- | --- | --- |
| 8 | 4147 | 1422 | 2.75x | **2.92x** |
| 24 | 5270 | 2096 | 2.47x | **2.51x** |
| 64 | 7259 | 3257 | 1.78x | **2.23x** |
| 256 | 18068 | 8827 | 1.40x | **2.05x** |
| 1024 | 49042 | 27324 | 1.09x | **1.79x** |
| 3000 | 124479 | 124306 | 1.00x | 1.00x |

`roaring_bitmap_or_inplace` moves 1.30x–2.28x over the same range.

Below 32 values/container the AVX-512 kernel is not used, so those rows are commit 1 alone. At 3000 values/container nothing changes because the union crosses the 4096-value threshold and converts to a bitset container. `or_many` is unchanged at every size — it goes through `lazy_or` with bitset conversion, so it never reaches this kernel.

### On the existing microbenchmarks

I ran the `microbenchmarks/bench` suite across four datasets and am deliberately **not** quoting it as evidence. It reported `SuccessiveUnion` at 0.91–0.94, but the same runs also moved `ContainsColdHigh` (0.67–0.69) and `AddOffset32771` (0.77–0.83), neither of which calls `fast_union_uint16` — so those runs carry layout/contention artifacts at a magnitude that swamps the signal. Two separate issues make that suite a poor instrument here, both worth addressing independently of this PR:

- `load_synthetic()` builds 30,000 bitmaps at startup, and at the default `--benchmark_min_time` that setup is ~37% of the process. The `binarySearch` (15–19%) and `__memmove_avx512` (10–16%) that top every profile are setup, not the kernel under test.
- `SuccessiveUnion` spends ~33% of its time in the allocator (`malloc_consolidate` 10%, `unlink_chunk` 9.4%, `_int_malloc` 8.2%), and its array containers are mostly under 32 values.

### Validation

- Correctness: 3078 randomized shapes against `union_uint16` (lengths 0..4096 including block boundaries 31/32/33, universes 200/4096/65536 so duplicate-heavy inputs are covered), 0 mismatches.
- Full `ctest` (26 suites) passes under gcc, clang, `ROARING_SANITIZE`, `ROARING_SANITIZE_UNDEFINED`, `-DROARING_DISABLE_AVX512=ON`, the amalgamation build, and arm64 (macOS).
- No new compiler warnings; formatted with clang-format 17 to match CI.

### Note on aliasing

`fast_union_uint16` is called with partially overlapping input and output from `array_run_container_inplace_union`. The new kernel preserves that contract the same way `union_vector16` does: after consuming `p1 + p2` whole blocks it has written at most `32*(p1+p2) - 32` values, because one merged block is always held back in `vmax`, so the write pointer stays strictly behind both read pointers. The tail does its two-way merges reading before writing.

### Follow-up, not in this PR

`vpternlogq` gives a 3-input bitset OR at **1.66x** over pairwise for 2 sources, 1.33x for 4, 1.14x for 8 — AVX2 has no equivalent. `roaring_bitmap_or_many` currently accumulates strictly pairwise via `lazy_or_inplace`, so it leaves that on the table. It needs a 3-input container-level op plus pairing in `or_many`, so it is a real change rather than a drop-in, but it is the thing that would move `TotalUnionHeap`.

### DIFF
--- a/include/roaring/array_util.h
+++ b/include/roaring/array_util.h
@@ -309,6 +309,17 @@ size_t union_uint32(const uint32_t *set_1, size_t size_1, const uint32_t *set_2,
 uint32_t union_vector16(const uint16_t *set_1, uint32_t size_1,
                         const uint16_t *set_2, uint32_t size_2,
                         uint16_t *buffer);
+
+#if CROARING_COMPILER_SUPPORTS_AVX512
+/**
+ * AVX-512 union of two sorted uint16 arrays, using a Batcher bitonic merge
+ * network over 32 lanes. Same contract as union_vector16: `buffer` must have
+ * room for size_1 + size_2 values.
+ */
+uint32_t avx512_union_uint16(const uint16_t *set_1, uint32_t size_1,
+                             const uint16_t *set_2, uint32_t size_2,
+                             uint16_t *buffer);
+#endif  // CROARING_COMPILER_SUPPORTS_AVX512
 /**
  * A fast SSE-based XOR function.
  */
--- a/include/roaring/array_util.h
+++ b/include/roaring/array_util.h
@@ -320,6 +320,7 @@ uint32_t avx512_union_uint16(const uint16_t *set_1, uint32_t size_1,
                              const uint16_t *set_2, uint32_t size_2,
                              uint16_t *buffer);
 #endif  // CROARING_COMPILER_SUPPORTS_AVX512
+
 /**
  * A fast SSE-based XOR function.
  */
--- a/src/array_util.c
+++ b/src/array_util.c
@@ -1746,6 +1746,171 @@ CROARING_UNTARGET_AVX2
  */
 
 /**
+ * Start of the AVX-512 16-bit union code.
+ *
+ * union_vector16 above merges 8 lanes at a time with an odd-even transposition
+ * network: 8 dependent min/max stages per 8 output values, which measures at
+ * ~2.3 cycles per input element -- no better than the scalar union_uint16.
+ *
+ * With AVX-512 we can merge 32+32 lanes with a Batcher bitonic network: one
+ * reverse plus 5 compare-exchange stages per sorted half, i.e. ~4x fewer
+ * operations per output value. This needs cross-lane 16-bit permutes (vpermw,
+ * vpermt2w) and a 16-bit compress store (vpcompressw); AVX2 has none of these,
+ * so the algorithm is genuinely AVX-512-only rather than a wider rerun of the
+ * SSE code.
+ */
+#if CROARING_COMPILER_SUPPORTS_AVX512
+
+CROARING_TARGET_AVX512
+
+// A compare-exchange at distance d pairs lane i with lane i^d and keeps the
+// smaller value in whichever lane has its d-bit clear. `hi` selects the lanes
+// whose d-bit is set.
+static inline __m512i avx512_cx16(__m512i v, __m512i t, __mmask32 hi) {
+    return _mm512_mask_mov_epi16(_mm512_min_epu16(v, t), hi,
+                                 _mm512_max_epu16(v, t));
+}
+
+// Sort a bitonic 32-lane sequence into ascending order. Each distance has a
+// dedicated cheap shuffle, so no stage needs a full vpermw.
+static inline __m512i avx512_bitonic_sort32(__m512i v) {
+    v = avx512_cx16(v, _mm512_shuffle_i64x2(v, v, 0x4E), 0xFFFF0000u);  // d=16
+    v = avx512_cx16(v, _mm512_shuffle_i64x2(v, v, 0xB1), 0xFF00FF00u);  // d=8
+    v = avx512_cx16(v, _mm512_shuffle_epi32(v, 0x4E), 0xF0F0F0F0u);     // d=4
+    v = avx512_cx16(v, _mm512_shuffle_epi32(v, 0xB1), 0xCCCCCCCCu);     // d=2
+    v = avx512_cx16(v, _mm512_rol_epi32(v, 16), 0xAAAAAAAAu);           // d=1
+    return v;
+}
+
+// Merge two ascending 32-lane vectors: *lo receives the 32 smallest values in
+// ascending order, *hi the 32 largest, also ascending.
+static inline void avx512_bitonic_merge32(__m512i a, __m512i b, __m512i *lo,
+                                          __m512i *hi) {
+    static const uint16_t revtab[32] = {
+        31, 30, 29, 28, 27, 26, 25, 24, 23, 22, 21, 20, 19, 18, 17, 16,
+        15, 14, 13, 12, 11, 10, 9,  8,  7,  6,  5,  4,  3,  2,  1,  0};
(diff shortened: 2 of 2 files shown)


lemire:
Thanks — the skew regression is real and I reproduced it independently before touching anything. Pushed a999279f.

You were right about the cause and right that my tables were blind to it: every shape I benchmarked had equal cardinalities, so nothing ever exercised the case where the small side holds fewer than two 32-lane blocks. There the block loop runs at most once and the rest falls to the scalar merge, while `union_vector16` keeps vectorizing at 8-lane granularity.

My reproduction, `roaring_bitmap_or` over 32 array containers, ns (`or` only, cardinalities under `DEFAULT_MAX_SIZE` so the result stays an array):

| card1/card2 | master | commit 1 only | PR before | PR now |
| --- | --- | --- | --- | --- |
| 16/1024 | 29689 | 26871 | 32113 | 27036 |
| 32/1024 | 28331 | 25323 | 34253 | 25830 |
| 40/1024 | 27910 | 24785 | 35479 | 25097 |
| 48/1024 | 28135 | 24994 | 36611 | 25392 |
| 64/1024 | 28911 | 25235 | 23823 | 20862 |
| 96/1024 | 28498 | 25524 | 20028 | 19235 |
| 40/4000 | 110198 | 106835 | 136510 | 107923 |
| 64/4000 | 109935 | 106484 | 104683 | 89701 |
| 96/4000 | 113242 | 109697 | 96919 | 86719 |
| 512/2048 | 57952 | 55217 | 33178 | 33894 |
| 2000/2000 | 106188 | 102847 | 68605 | 68605 |

1.29x slower at 48/1024 and 1.46x against commit 1 alone — matching your numbers closely. With both gated fallbacks every shape is now at or above master, and the ones that were already winning improve, since the tail vectorizes too: 64/4000 goes 1.05x → 1.23x, 96/4000 1.17x → 1.31x. Worst case anywhere is 128/4000 at 0.99x, which is noise. Balanced n-vs-n is unchanged at every size (`orbench` identical to the previous push within 2%).

I took your gating thresholds as-is; they land in the right place in my measurements too.

**Revalidation.** I rebuilt the correctness harness around the boundaries this change introduces — sizes stepping through every block edge and both 64 thresholds, universes 64/200/4096/65536, disjoint / strictly interleaved / identical inputs, both argument orders, the `fast_union_uint16` dispatch, and the aliased in-place layout from `array_run_container_inplace_union` — with every output buffer `malloc`'d at exactly `l1+l2` so an overrun trips. 125,648 cases, 0 failures, under ASAN+UBSAN. Full ctest 26/26 under gcc, clang, ASAN+UBSAN, `ROARING_DISABLE_AVX512=ON`, `ROARING_DISABLE_AVX=ON`, amalgamation, and arm64. No new warnings.

**On the minor notes.** Took the tightness comment — it now says explicitly that the bound is exactly tight and that shrinking what the tail holds back breaks aliased callers silently. Took the blank line.

I left the `#if CROARING_COMPILER_SUPPORTS_AVX512` guard alone. You're right that the macro is only defined inside `#if CROARING_IS_X64` in `isadetection.h` and so is undefined off-x64, but line 226 of the same header already does exactly this, as do the `#ifndef ... #error` blocks at the top of `array_util.c`, `bitset_util.c`, `array.c`, `bitset.c`, `convert.c` and `run.c`. Fixing only my line would leave the file inconsistent without making `-Wundef` viable. Happy to do the sweep as a separate PR if you want it.

Also agreed on `sort_uint16_short` — `leftoversize` is 8 plus at most 7, so ≤15; "at most 16" is a correct upper bound and insertion sort is fine there.

PR #859, opened 2026-08-06
Fix non-atomic refcount decrement

### LINKED ISSUE #858: Non-atomic refcount decrement never reports zero
### Description

On current `master` (`b3146bd2`), `croaring_refcount_dec()` compares the pointer rather than the decremented value when `CROARING_ATOMIC_IMPL_NONE` is selected:

```c
*val -= 1;
return val == 0;
```

Because callers pass a valid pointer, the function never reports that the last reference was released. `shared_container_free()` consequently does not free the shared container, and `shared_container_extract_copy()` takes its copy path even when the reference count reached zero.

### Reproduction

```c
#include <assert.h>
#include <roaring/portability.h>

int main(void) {
    croaring_refcount_t refcount = 1;
    assert(croaring_refcount_dec(&refcount));
}
```

Compile and run against the repository root:

```sh
cc -std=c99 -DCROARING_ATOMIC_IMPL=CROARING_ATOMIC_IMPL_NONE -Iinclude refcount-repro.c
./a.out
```

The assertion fails. The expected result is `true` after decrementing the count from one to zero.

I reproduced this on Ubuntu with GCC 13.3.0. The issue is specific to the supported non-atomic fallback selected explicitly above; the C, C++, and Windows atomic implementations already compare the decremented value correctly.

I found this while reviewing the vendored CRoaring integration in redis/redis#15331. I am preparing a minimal fix with a regression test that forces the non-atomic implementation.

aviggiano:
## Summary

- compare the decremented reference count, rather than its pointer, in the non-atomic fallback
- add a dedicated regression test that forces `CROARING_ATOMIC_IMPL_NONE`

## Root cause and impact

`croaring_refcount_dec()` decremented `*val` but returned `val == 0`. Since callers pass a valid pointer, the function could never report that the final reference had been released. This prevents shared containers from being freed and makes extraction take an unnecessary copy path when the non-atomic fallback is selected.

Fixes #858.

This was found while reviewing the vendored CRoaring integration in redis/redis#15331.

## Validation

- confirmed the new regression test fails on unmodified `master`
- `cmake -S . -B build-release -DCMAKE_BUILD_TYPE=Release -DENABLE_ROARING_TESTS=ON`
- `cmake --build build-release -j2`
- `ctest --test-dir build-release --output-on-failure` — 27/27 passed
- forced-fallback test compiled as C++ with Clang 18 and passed
- `git diff --check`

The Docker-based formatting helper was unavailable locally; formatting remains covered by CI.

Prepared with Codex assistance under contributor direction; the bug was reproduced and the complete test suite above was run before opening this draft.


lemire:
Seems like an obvious defect to me.  Merging.

PR #879, opened 2026-09-11
armv7 FreeBSD fixes

clausecker:
This fixes two portability issues found on armv7 and FreeBSD.

 - the `bswap64()` macro requires `<sys/endian.h>` to be included
 - `vmaxvq_u16()` is not available on 32-bit ARM

See the two commits for details.


clausecker:
On another note, `vmaxvq_u16()` is suboptimal for checking for zero.  It's probably faster to reduce once with `vpmaxq_u16()` into 4 16-bit integers, which you then compare to equality with  zero.  This is because the `umaxv` instruction on `8h` organisation takes 5 cycles, throughput of 1 per cycles, whereas the `umaxp` instruction on `8h` organisation is much cheaper at 2 cycles, throughput 2 per cycle.

PR #882, opened 2026-09-11
faster zero check on ARM

lemire:
Follow-up to https://github.com/RoaringBitmap/CRoaring/pull/879 by @clausecker


clausecker:
This unfortunately doesn't work if the FZ (denormals are zero) flag is set, in which case some bit patterns compare equal to floating point zero that are not expected to.  I don't think you are in control of floating-point settings in the context of this library.

You can play around with it with [this test program](https://gist.github.com/clausecker/723ed29d7af4d6a79da1b9ced28e2350) I wrote in the context of [this Stack Overflow question](https://stackoverflow.com/q/79336641/417501).  I think we've discussed this earlier as well.

lemire:
You are, of course, correct. It is even explained on my blog.

PR #885, opened 2026-09-17
Fix clang-format after macro renaming

anemitz:
Formats the renamed assembly macros and wraps the lazy OR call to satisfy clang-format 17.

Fixes the failed clang-format CI job: https://github.com/RoaringBitmap/CRoaring/actions/runs/35254981990


anemitz:
@lemire cleanup for naming changes introduced in #844

PR #888, opened 2026-09-22
Avoid forming misaligned typed pointers in (de)serialization

lemire:
## Summary

Two places formed misaligned typed pointers that were only ever passed to `memcpy`:

- `roaring64_bitmap_frozen_serialize` accepts an unaligned output buffer (only `roaring64_bitmap_frozen_view` requires 64-byte alignment), but it carved the output into `uint64_t *` / `uint16_t *` / `rle16_t *` cursors aligned *relative to the buffer start*. With an unaligned base pointer (as `check_frozen_serialization` in `roaring64_unit` deliberately uses), those typed pointers are misaligned.
- `roaring_bitmap_deserialize` and `roaring_bitmap_deserialize_safe` formed `const uint32_t *elems` at `buf + 5` and computed `elems + i` before the `memcpy`. The comment already said the data may be unaligned; the pointer type just didn't match that intent.

Nothing faults at run time since the copies go through `memcpy`, but forming a misaligned pointer to a type is undefined behavior (C11 6.3.2.3p7). clang's `-fsanitize=alignment` checks typed `memcpy` arguments and reports both; GCC's UBSan (what `ubuntu-sani-ub-ci` uses) does not, which is why CI is green.

The fix keeps byte (`char *`) cursors and advances them by byte counts. `container_get_frozen_size()` already returns the byte count copied, so `container_frozen_serialize` now advances by it directly. Serialized output is byte-for-byte unchanged.

## Test plan

- [x] clang + `-DROARING_SANITIZE=ON -DROARING_SANITIZE_UNDEFINED=ON`: 27/27 tests pass, zero `runtime error` lines in the ctest log (previously `roaring64_unit` reported a misaligned `uint16_t *` store in `container_frozen_serialize`, and `toplevel_unit` aborted in `test_serialize` at `roaring.c:1687`)
- [x] Release build: 27/27 tests pass
- [x] clang-format 17 clean


Dr-Emann:
It _always_ seems crazy that just _creating_ an unaligned pointer is UB in C, something that always catches me when going from rust to C/C++.

lemire:
@Dr-Emann I don't think it actually matters. I am being pedantic here. I would not fix it if not for the sanitizer warnings.

